PR #8, opened 2018-07-31
Add starting-point for python support

jokva:


JensGM:
Do we need CMake at all?

jokva:
It's not _necessary_, but it's convenient to be able to run all from cmake (in my opinion) so that it *can* be a single-step operation to build.

jokva:
Oh and it's nice documentation on how to build too.

jokva:
Replaced by #9

PR #75, opened 2019-03-28
io.cpp positive tests and remaining protocol.cpp tests

achaikou:


lib/src/io.cpp
```diff
@@ -303,6 +309,7 @@ record& stream::at( int i, record& rec ) noexcept (false) {
 
             if (err) consistent = false;
             attributes.push_back( attrs );
+            types.push_back( type );
```
jokva: What's going on here?

python/dlisio/ext/core.cpp
```diff
@@ -418,10 +418,11 @@ PYBIND11_MODULE(core, m) {
...
             return b;
         })
-        .def( "extract", [](dl::stream& s, const std::vector< long long >& tells) {
+        .def( "extract", [](dl::stream& s,
+                            const std::vector< long long >& tells_indices) {
             std::vector< dl::record > recs;
-            recs.reserve( tells.size() );
-            for (auto i : tells) {
+            recs.reserve( tells_indices.size() );
+            for (auto i : tells_indices) {
```
jokva: I'm not sure this is actually a clarification. If it's indices anyways, why mix tells into it?

achaikou: I agree, it actually is just "indices" to extract, which would be more clear. But for some reason it has signature of "std::vector< long long >" (which corresponds to tells one) and I am unable to understand where we got it from. 
If that's not a problem, I would happily change it just to "indices".

lib/src/io.cpp
achaikou: Every record has a "type" field. There is a line further in the method that reads type from types vector (types.front()). But we never put current type there, so it was always reading "0" from empty vector before.

PR #82, opened 2019-04-03
pack.cpp & Co

achaikou:


lib/src/dlisio.cpp
```diff
@@ -667,19 +667,27 @@ int dlis_pack_varsize(const char* fmt, int* src, int* dst) {
...
-    int short_correction = 0;
+    int src_dst_correction = 0;
+    auto mem_size = 0;
     int size = 0;
     while (true) {
         switch (*fmt++) {
             case DLIS_FMT_EOL:
-                if (varsrc) short_correction = size;
-                if (src) *src = size - short_correction;
+                if (varsrc) src_dst_correction = size;
```
jokva: Rename please.

lib/src/dlisio.cpp
```diff
@@ -667,19 +667,27 @@ int dlis_pack_varsize(const char* fmt, int* src, int* dst) {
...
+                if (src) *src = size - src_dst_correction;
                 if (dst) *dst = size;
                 return DLIS_OK;
 
             case DLIS_FMT_FSHORT:
-                short_correction += 2;
-                size += sizeof(float);
+                mem_size = sizeof(float);
```
jokva: mem time is unnecessary, please remove

lib/test/pack.cpp
```diff
@@ -11,17 +11,42 @@
 
 namespace {
 
+void check_size(const char* fmt, const int* src_v_size, const int* dst_v_size){
```
jokva: check_size shouldn't be a separate function, and/or should not look like this with pointer shenanigans. _Intent_

jokva:
Squash please.

PR #83, opened 2019-04-05
Find sul in core

jokva:


lib/src/dlisio.cpp
```diff
@@ -258,6 +258,29 @@ int dlis_sul( const char* xs,
     return DLIS_UNEXPECTED_VALUE;
 }
 
+int dlis_find_sul(const char* from,
+                  long long search_limit,
+                  long long* offset) {
+
+    static const auto needle = "RECORD";
+    const auto to = from + search_limit;
```
achaikou: I would suggest to add to docs that if someone passes search limit that reads outside of memory, it's their problem.

lib/src/dlisio.cpp
```diff
@@ -258,6 +258,29 @@ int dlis_sul( const char* xs,
...
+    const auto itr = std::search(from, to, needle, needle + 6);
+
+    if (itr == to)
+        return DLIS_NOTFOUND;
+
+    /*
+     * Before the structure field of the SUL there should be 10 bytes, i.e.
```
achaikou: I think it should be 9 bytes

lib/test/sul.cpp
```diff
@@ -383,3 +383,62 @@ TEST_CASE_METHOD(correct_sul, "SUL with garbage minor/major version", "[sul]") {
...
+
+TEST_CASE("Find SUL when there is no garbage", "[sul]") {
+    long long off = -1;
+    const auto err = dlis_find_sul(plain_sul, sizeof(plain_sul), &off);
+
+    CHECK(!err);
+    CHECK(off == 0);
+}
+
+TEST_CASE("find_sul gracefully handles missing SUL", "[sul]") {
```
achaikou: To additionally guard against off-by-1 possible issue, it might be worth using test
"   1V1.00RECORgarbage....    "

lib/src/dlisio.cpp
jokva: I considered it, but's given by having a search_limit and a pointer in the interface, so I omitted it.

lib/src/dlisio.cpp
jokva: Yes.

PR #97, opened 2019-04-22
Re-format dtype.names on duplicate mnemonics

### LINKED ISSUE #95: Better dealing with duplicated mnemonics when retrieving log values
When I try to load the curve values from a frame with repeated mnemonics, something like:

> `fingerprint = fr.fingerprint # A frame fingerprint`
> `curves = f.curves(fingerprint) # Load all curves`

I get the following error: `ValueError: field 'SPT4' occurs more than once`. SPT4 may be any duplicated mnemonic.

This kind of issue is really common with DLIS files. I suggest using the [same approach from lasio](https://lasio.readthedocs.io/en/latest/_modules/lasio/las_items.html),  any duplicates will have ':1', ':2', ':3', etc., appended to them. 

That way the API user can inspect the data and decide what which logs he really wants to work with.

I'm using DLISIO version 0.1.4, installed directly using pip.

jokva:
When a frame has multiple channels with the same name, enrich the label
by adding :origin:copynumber, so that all names given to numpy.dtype are
unique. This is already the built-in disambiguation mechanism of RP66.

This addition is not done by default because for the cases with only one
mnemonic for the frame, the added formatting is obnoxious and
unnecessary.

Closes #95

### DIFF
--- a/python/dlisio/plumbing/basicobject.py
+++ b/python/dlisio/plumbing/basicobject.py
@@ -21,8 +21,8 @@ class BasicObject():
 
         try:
             self.name       = name.id
-            self.origin     = name.origin
-            self.copynumber = name.copynumber
+            self.origin     = int(name.origin)
+            self.copynumber = int(name.copynumber)
         except AttributeError:
             pass
 
--- a/python/dlisio/plumbing/frame.py
+++ b/python/dlisio/plumbing/frame.py
@@ -3,6 +3,7 @@ from ..reprc import fmt
 from .valuetypes import scalar, vector, boolean
 
 import numpy as np
+import logging
 
 
 class Frame(BasicObject):
@@ -81,6 +82,15 @@ class Frame(BasicObject):
         data-type of each frame. I.e. the sum of channel.dtype of each channel
         in self.channels.
 
+        If all curve mnemonics are unique, then dtype.names == [ch.name for ch
+        in self.channels]. If there are more than one channel with the same
+        name for this frame, all duplicated mnemonics are enriched with origin
+        and copynumber.
+
+        Consider a frame with the channels mnemonics [('TIME', 0, 0), ('TDEP',
+        0, 0), ('TIME, 1, 0)]. The dtype names for this frame would be
+        ('TIME:0:0', 'TDEP', 'TIME:1:0').
+
         See also
         --------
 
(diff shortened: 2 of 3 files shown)


rmorel:
Hey @jokva I tested your solution. But I'm still getting a ``ValueError`` exception.

For example:

```
f = dlisio.load(test_file)
fingerprint = 'T.FRAME-I.B414452-O.0-C.0' # A valid fingerprint of the frame w/ repeated mnemonics
f.curves(fingerprint)
```

This gives me the following error message:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-893653c07260> in <module>()
      2 f = dlisio.load(test_file)
      3 fingerprint = 'T.FRAME-I.B414452-O.0-C.0' # A valid fingerprint of the frame w/ repeated mnemonics
----> 4 f.curves(fingerprint)

C:\ProgramData\Anaconda3\lib\site-packages\dlisio\__init__.py in curves(self, fingerprint)
    125         indices = self.fdata_index[fingerprint]
    126         a = np.empty(shape = len(indices), dtype = frame.dtype)
--> 127         core.read_all_fdata(fmt, self.file, indices, a)
    128         return a
    129 

ValueError: ':' is not an allowed character in buffer field names
```

I'm guessing this is a limitation from numpy, maybe you can choose another character instead of ':'. I'll do some further tests and investigate the ``frame.dtype.names`` property.

jokva:
Interesting. The test for this particular feature uses the : character, and it works on all installations we run CI on. What python and numpy versions are you using?

jokva:
Actually, the problem isn't the dtype (which supports it), the problem is then passing it as a buffer object.

```
>>> import numpy as np
>>> dtype = np.dtype([('TIME:0', 'i4')])
>>> dtype
dtype([('TIME:0', '<i4')])
>>> xs = np.array(10, dtype = dtype)
>>> xs.dtype
dtype([('TIME:0', '<i4')])
>>> memoryview(xs)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: ':' is not an allowed character in buffer field names
```

jokva:
Fortunately, that makes the solution rather simple - we replace : with . (and make sure the tests calls `memoryview` on the array)

rmorel:
Good idea. I tested replacing manually all ':' with '.' in ``frame._dtype`` and I was able to read the curve values afterwards.

Since you asked, I'm using numpy 1.14.3 on Windows 10.

jokva:
Now it doesn't matter that much anymore, but good to know should there be more issues.

jokva:
@rmorel We just published 0.1.6 which contains this fix.

rmorel:
@jokva I just tested with DLISIO 0.1.6. Everything works fine now. Thanks for the help.

jokva:
Excellent, thanks for the report and help testing this feature.

PR #108, opened 2019-05-14
Change linkage

achaikou:


python/dlisio/plumbing/linkage.py
```diff
@@ -0,0 +1,7 @@
+def obname(objtype):
+    def fingerprint(obj):
```
ErlendHaa: Raise TypeError if not obname

python/dlisio/plumbing/linkage.py
```diff
@@ -0,0 +1,7 @@
...
+        return obj.fingerprint(objtype)
+    return fingerprint
+
+def objref(obj):
```
ErlendHaa: Raise TypeError if not fingerprint

python/dlisio/plumbing/basicobject.py
```diff
@@ -126,13 +134,48 @@ def strip(d):
...
+                ref = self.refs[attr]
+            except KeyError:
+                continue
+
+            if ref is None: continue
+
+            def get_link(v):
+                try:
+                    fp = link(v)
+                except (AttributeError, TypeError):
```
ErlendHaa: Is AttributeError needed?

ErlendHaa: Also, add  msg to logging.warning! This should only happen if someone adds a none obname/objref to self.refs before link/relink.

python/dlisio/plumbing/basicobject.py
```diff
@@ -14,6 +16,7 @@ class BasicObject():
     """
 
     attributes = {}
+    linkage    = {}
```
ErlendHaa: Maybe a docstring explaning the intentions of linkage and how it can be patch

python/dlisio/plumbing/basicobject.py
```diff
@@ -17,6 +17,19 @@ class BasicObject():
 
     attributes = {}
     linkage    = {}
+    """dict: Shows how to to interpret link objects stored in refs.
+    During load stage all the objects of obname, objref and attref types
+    are put into refs. During link stage they are linked to actually
+    loaded objects based on information provided in linkage.
+    If linkage is to be updated manually for the class:
+
+    >>> Coefficient.linkage['myparam'] = linkage.obname("PARAMETER")
```
ErlendHaa: Maybe specify that the values **must** be either `linkage.obname("TYPE")` or `linkage.objref`?

python/dlisio/plumbing/basicobject.py
```diff
@@ -17,6 +17,19 @@ class BasicObject():
 
     attributes = {}
     linkage    = {}
+    """dict: Shows how to to interpret link objects stored in refs.
```
ErlendHaa: Could maybe be a bit clearer. "Linkage defines the rules for how each attribute links to other objects"?

python/dlisio/plumbing/basicobject.py
```diff
@@ -16,6 +16,21 @@ class BasicObject():
     """
 
     attributes = {}
+    """dict: Maps dlis-file attributes to class attributes.
```
ErlendHaa: Add something like this to clearly state the intent of attribute: "Index of what attributes this object type is expected to have. dlisio will use this index when parsing object. If an attribute is not in this index, it will be ignored when parsing the file"?

PR #109, opened 2019-05-15
Add more logical record types to Python

ErlendHaa:


python/dlisio/plumbing/longname.py
```diff
@@ -0,0 +1,87 @@
...
+
+        #: Conditions applicable at the time the information was acquired or
+        #: generated
+        self.conditions      = []
+
+        #: Industry-standardized symbolic name by which the information is
+        #: known. The possible values are specified by POSC
+        self.standard_symbol = None
+
+        #:
```
achaikou: Maybe "Association between the recorded information and corresponding records or objects of the Producer’s internal or corporate database"?

python/tests/test_semantics.py
```diff
@@ -75,6 +75,34 @@ def test_origin(f):
     assert random_origin.file_set_nr == 1042
     assert random_origin.file_nr     == 6
 
+def test_axis(f):
```
achaikou: Btw, I still believe that core tests on actual files might be beneficial. Maybe there is no need to test every single attribute, but checking number of objects or whatever might still provide some value.

ErlendHaa: I have looked through all of the test files, and none of them contains objects such as long_name, axis or zone! Maybe we can get some additional testfiles in the future

python/tests/test_semantics.py
```diff
@@ -75,6 +75,34 @@ def test_origin(f):
...
+    assert ln.entity_nr       == '21'
+    assert ln.entity_part     == 'top part?'
+    assert ln.entity_part_nr  == '1'
+    assert ln.generic_source  == 'random words'
+    assert ln.source_part     == ['generator']
+    assert ln.source_part_nr  == ['8412']
+    assert ln.conditions      == ['at standard temperature']
+    assert ln.standard_symbol == 'SYM'
+    assert ln.private_symbol  == 'BOL'
+
```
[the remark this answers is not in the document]
ErlendHaa: Agree! We need to update all links to these new objects from the existing ones, but it makes sense to wait for the new link-routine.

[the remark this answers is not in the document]
ErlendHaa: if #108 is merged first, I will update linking for these and existing objects! With tests ofc :D

python/dlisio/plumbing/computation.py
```diff
@@ -0,0 +1,57 @@
...
+        #: computation is constant
+        self.zones       = []
+
+        #: Computational values
+        self.values      = []
+
+        #: The immediate source of the Computation
+        self.source = []
+
+        #: Reference to the source object
```
achaikou: You shouldn't need it

Issue #111, opened 2019-05-20
Possible regression on dlisio.load()


rmorel:
When I try to load many of my DLIS files on DLISIO 0.1.9 as (for example):
```
import dlisio
test_file = r'R:\3-BRSA-1053-RJS\PERFIS_DIGITAIS\DLIS\3-brsa-1053-rjs_8aittdd_dsi-7845.dlis'
f = dlisio.load(test_file)
```

I get the following error:

```
ERROR:root:multiple distinct objects in set FILE-HEADER (). Duplicate fingerprint = T.FILE-HEADER-I.1-O.35-C.0
WARNING:root:continuing with the last object
ERROR:root:multiple distinct objects in set ORIGIN (). Duplicate fingerprint = T.ORIGIN-I.DLIS_DEFINING_ORIGIN-O.35-C.0
WARNING:root:continuing with the last object
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-0faac13e171f> in <module>()
      1 test_file = r'R:\3-BRSA-1053-RJS\PERFIS_DIGITAIS\DLIS\3-brsa-1053-rjs_8aittdd_dsi-7845.dlis'
----> 2 f = dlisio.load(test_file)

C:\ProgramData\Anaconda3\lib\site-packages\dlisio\__init__.py in load(path)
    334     try:
    335         stream.reindex(tells, residuals)
--> 336         f = dlis(stream, explicits, sul_offset = sulpos)
    337 
    338         explicits = set(explicits)

C:\ProgramData\Anaconda3\lib\site-packages\dlisio\__init__.py in __init__(self, stream, explicits, sul_offset)
     40             'COMPUTATION'            : plumbing.Computation.create,
     41         }
---> 42         self.load()
     43 
     44     def __enter__(self):

C:\ProgramData\Anaconda3\lib\site-packages\dlisio\__init__.py in load(self, sets)
    103 
    104                     logging.info(duplicate.format(fingerprint))
--> 105                     if original.attic != obj.attic:
    106                         msg = problem + where
    107                         msg = msg.format(os.type, os.name, fingerprint)

TypeError: __eq__(): incompatible function arguments. The following argument types are supported:
    1. (self: dlisio.core.objref, arg0: Tuple[str, Tuple[int, int, str]]) -> bool

[...]

PR #113, opened 2019-05-21
Register operators != and == for core.objref

ErlendHaa:
dl::objref operators != and == was not registered in the python
extension. dlis.load() uses these operators when duplicated fingerprints
are found, as a result the entire load procedure failed.


ErlendHaa:
Fixes #111

PR #115, opened 2019-05-27
Fix travis build

achaikou:
Travis build failed.
Leading theory - new cmake depends on scikit-build.
Hence trying to install it before installing cmake.


jokva:
Fixed in #116.

PR #129, opened 2019-06-13
Padbytes & logical files fixes

### LINKED ISSUE #121: Bad segment trims
Example: 
`padbytes (which is 168) >= segment.size() (which is 136)`

### LINKED ISSUE #123: Multiple frames in one FDATA
Not implemented yet: multiple frames in one FDATA

achaikou:
closes #121 
closes #123 

+ a few commits of minor interest


python/dlisio/reprc.py
```diff
@@ -59,13 +59,13 @@
...
-    23     : 'i4,i1,i1,U',           #Object name
-    24     : 'i1,U,i4,i1,i1,U',      #Object reference
-    25     : 'i1,U,i4,i1,i1,U,i1,U', #Attribute reference
-    26     : 'i1',                   #Boolean status
-    27     : 'i1,U',                 #Units expression
+    22     : 'u4',                   #Origin reference
+    23     : 'u4,u1,u1,U',           #Object name
+    24     : 'u1,U,u4,u1,u1,U',      #Object reference
+    25     : 'u1,U,u4,u1,u1,U,u1,U', #Attribute reference
+    26     : '?',                    #Boolean status
```
jokva: What's this question mark?

python/dlisio/ext/core.cpp
```diff
@@ -605,7 +605,11 @@ PYBIND11_MODULE(core, m) {
             recs.reserve( indices.size() );
             for (auto i : indices) {
                 auto rec = s.at( i );
-                if (rec.isencrypted()) continue;
+                if (rec.isencrypted()){
+                    /* as we don't want to add encrypted data yet, adding
+                       empty record instead to preserve data structure */
+                    rec.data.clear();
```
jokva: I think this isn't a great fix.

If we see the record is encrypted we should probably be able to return all the bytes intact, and leave dealing with padbytes to whoever's decrypting it. With this, we now have to deal with encrypted records _and_ we destroy the data in the process.

achaikou: I fail to see how it is different from what we had before? I wanted to avoid changing current logic.
Previously if record was encrypted we as well just ignored the record, so it was as is we never added the data. Right now we add placeholder instead.
Do you have a better suggestion for these records?

python/dlisio/reprc.py
achaikou: It's boolean. (docs.scipy seem to be down, so cached link)
http://webcache.googleusercontent.com/search?q=cache:tyoWStj_4OcJ:https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.dtypes.html&hl=en&gl=no&strip=1&vwsrc=0
Tests for currently supported representation codes are coming in other PR.

python/dlisio/ext/core.cpp
jokva: >  I wanted to avoid changing current logic.

Don't, current logic is awful.

> I fail to see how it is different from what we had before?

I suppose it's not too different, except we're now still signaling an error. It's different because if the record is encrypted we don't even try to trim, whereas before we always trimmed. When encrypted, we should probably keep on going, return the byte array, but flag the fact that it's rubbish until decrypted.

> Previously if record was encrypted we as well just ignored the record, so it was as is we never added the data.

Sure, and this is fine. There should be a recovery though, for the cases where the caller _can_ decrypt.

python/dlisio/reprc.py
jokva: Sure, but the logical/boolean in DLIS has a given width, will this always line up? Probably not, and it's quite the risky assumption to make.

achaikou: Numpy uses 1 byte for booleans, as well as DLIS. I doubt this is going to change, so I don't see any risk here. Numpy also doesn't accept size specifier near '?'.
I am not fine with user getting integer value in place of boolean, and this seems the best place to make the conversion.

python/dlisio/ext/core.cpp
achaikou: Changed for encrypted records not to be treated any differently.
User has to check record.isencrypted() to find out if this record is garbage or not.
If I still misunderstand what you are communicating, lets talk through slack.

python/dlisio/reprc.py
jokva: You're right, then it makes perfect sense.

PR #152, opened 2019-08-23
Indexedobjects

achaikou:
A bit different from initial agreements (channels/frames & stuff properties; objects property; object method), so we can discuss.


python/dlisio/__init__.py
```diff
@@ -325,6 +323,24 @@ def object(self, type, name, origin, copynr):
             msg = "Object {}.{}.{} of type {} is not found"
             raise ValueError(msg.format(name, origin, copynr, type))
 
+    @property
```
ErlendHaa: I don't see why this can't just be an attribute as before?

python/dlisio/__init__.py
```diff
@@ -365,8 +365,11 @@ def object(self, type, name, origin, copynr):
         type: type as specified in RP66
         name: name
         origin: origin number
+                May be None. If origin or copynumber not provided, search among
```
ErlendHaa: All parameters should be stated like this: 
``` 
name : type
    description
```

python/dlisio/__init__.py
```diff
@@ -379,12 +382,27 @@ def object(self, type, name, origin, copynr):
...
+                return obj[0]
+            elif len(obj) == 0:
+                msg = "No objects with name {} and type {} are found"
+                raise ValueError(msg.format(name, type))
+            elif len(obj) > 1:
+                msg = "Found several objects with requested parameters: \n{}"
+                desc = ""
+                for o in obj:
+                    desc += ("Object {} with origin {} and copynumber {}. \n"
+                             .format(o, o.origin, o.copynumber))
```
ErlendHaa: I think this warning have the potential to be very long with the current format. what about: 
There are multiple <_type_>'s named <_name_>. Found: (origin=1, copy=1), (origin=1, copy=2), (origin=2, copy=1)

python/dlisio/__init__.py
```diff
@@ -92,7 +92,25 @@ def __init__(self, t):
...
-    fileheader   = IndexedObjectDescriptor("FILE-HEADER")
+    @property
+    def fileheader(self):
+        """ Read Fileheader object
+
+        Returns
+        -------
+        fileheader : dict_values
+
+        """
```
ErlendHaa: I don't like that all this stuff is above __ init __, consider moving it?

PR #158, opened 2019-09-24
TDD for expected fdata fixes

achaikou:
Some screw-ups possible, so if during actual code development something would behave weirdly, let me know.


achaikou:
I really hope that it's now easy to read, investigate and check for errors at least for someone.

PR #159, opened 2019-09-25
Update existing docs and add some welcoming words to sphinx

### LINKED ISSUE #156: Update the sphinx documentation
The sphinx docs could use some improments.

- An introduction to the project
- Some concepts of dlis cannot be abstracted away by dlisio, hence should be explained
-  A quick guide to get started with using dlisio. Keep an emphasis on referring to the module docstrings for more detailed documentation. It is preferable that these are the main source of documentation as they are way easier to keep up to date. 
- General improvements for the existing docs 

Some of this have deliberately been postponed because we keep breaking the interface with new changes, but now seems a good time to at least give some sort of guide to how to use dlisio

ErlendHaa:
Closes #156

### DIFF
--- a/python/dlisio/__init__.py
+++ b/python/dlisio/__init__.py
@@ -500,8 +500,8 @@ def load(path):
     Notes
     -----
 
-    1) That the parentezies are needed when unpacking directly in the with
-    statment
+    1) That the parentheses are needed when unpacking directly in the with
+    statement
 
     2) The asterisk allows an arbitrary number of extra logical files to be
     stored in tail. Use len(tail) to check how many extra logical files there
--- a/python/dlisio/__init__.py
+++ b/python/dlisio/__init__.py
@@ -13,6 +13,44 @@ except pkg_resources.DistributionNotFound:
     pass
 
 class dlis(object):
+    """Logical file
+
+    A logical file is a collection of objects - in lack of a better word,
+    that are logically connected in some way. Think of a logical file as a pool
+    of objects. Some examples of objects are Channel, Frame, Fileheader, Tool.
+    They are all accessible through their own attribute.
+
+    :py:func:`object()` and :py:func:`match()` let you access specific objects
+    or search for objects matching a regular expression. Unknown objects such
+    as vendor-specific objects are accessible through the 'unknown'-attribute.
+
+    Uniquely identifying an object within a logical file can be a bit
+    cumbersome, and confusing.  It requires no less than 4 attributes!  Namely,
+    the object's type, name, origin and copynumber.  The standard allows for
+    two objects of the same type to have the same name, as long as either their
+    origin or copynumber differ. E.g. there may exist two channels with the
+    same name/mnemonic.
+
+    Attributes
+    ----------
+
+    attic : dict
+        Primitive dict-like representations of the objects. These objects are
+        the output of the parsing routine of the core library. They are the
+        foundation of which the more rich python objects are created from.
+        They are considered to be an internal data structure, the typical user
+        does not need to care about these.
+
+    problematic : list
+        Duplicated objects. If dlisio is not able to uniquely identify an object
+        based on the standards definition, the object is flagged as
+        problematic. Mainly indended for debugging files.
+
+    indexedobject : dict
+        A full inventory of all objects in the logical file, indexed by type.
+        Note that there are more handy ways of accessing objects than through
+        this dictionary.
+    """
     types = {
         'AXIS'                   : plumbing.Axis,
         'FILE-HEADER'            : plumbing.Fileheader,
@@ -37,11 +75,14 @@ class dlis(object):
         'COMMENT'                : plumbing.Comment,
     }
 
-    """dict: Parsing guide for native dlis object-types. Maps the native dlis
-    object-type to python class. E.g. all dlis objects with type
-    AXIS will be constructed into Axis objects. It is possible to both remove
-    and add new object-types before loading or reloading a file. New objects
-    will behave the same way as the defaulted object-types.
+    """dict: Parsing guide for native dlis object-types. The typical user can
+    safely ignore this attribute. It is mainly intended for internal use and
+    advanced customization of dlisio's parsing routine.
+
+    Maps the native dlis object-type to python class. E.g. all dlis objects
+    with type AXIS will be constructed into Axis objects. It is possible to
+    both remove and add new object-types before loading or reloading a file.
+    New objects will behave the same way as the defaulted object-types.
 
     If an object-type is not in types, it will be default parsed as an unknown
     object and accessible through the dlis.unknowns property.
@@ -79,8 +120,10 @@ class dlis(object):
     See also
     --------
 
-    plumbing.BasicObject.attributes : attributes
-    plumbing.BasicObject.linkage    : linkage
+    plumbing.BasicObject.attributes : Attributes of an object can be customized
+      in a similar manner as types.
+    plumbing.BasicObject.linkage    :  Linkage tells dlisio how it should try
+       to link the content of attributes to other objects.
     """
 
     def __init__(self, stream, explicits, attic, implicits, sul_offset = 80):
@@ -102,6 +145,12 @@ class dlis(object):
         self.close()
 
     def close(self):
+        """Close the file handle
+
+        It is not necessary to call this method if you're using the `with`
+        statement, which will close the file for you. Calling methods on a
+        previously-closed file will raise `IOError`.
+        """
         self.file.close()
 
     def __repr__(self):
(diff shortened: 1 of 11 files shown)


python/docs/examples.rst
```diff
@@ -0,0 +1,203 @@
...
+    
+    >>> with dlisio.load(filename) as files:
+    ...     for f in files:
+    ...         pass
+
+The returned :code:`files` can be iterated over and operations can be applied
+to each logical file.
+
+If you only want to work with one logical file at the time,
+:py:func:`dlisio.load()` support automatic unpacking of logical file. The
```
achaikou: support(s) ...
and maybe 
unpacking of logical file(s)
looks better

python/docs/examples.rst
```diff
@@ -0,0 +1,203 @@
...
+Accessing objects
+-----------------
+
+Think of :ref:`Logical files` as pools of objects with different types.  All
+objects of a type can be reached by name:
+
+.. code-block:: python
+
+    >>> for ch in f.channels
+    ...     _ = ch
```
achaikou: Weird action as for a quickstart example.

python/docs/examples.rst
```diff
@@ -0,0 +1,203 @@
...
+    --
+    440-CHANNEL        : 95
+    440-OP-CORE_TABLES : 250
+    440-OP-CHANNEL     : 93
+
+Accessing objects
+-----------------
+
+Think of :ref:`Logical files` as pools of objects with different types.  All
+objects of a type can be reached by name:
```
achaikou: Wasn't clear to me what you meant, so maybe
All objects of a type can be reached by name, for example channels or coefficients:

python/docs/examples.rst
```diff
@@ -0,0 +1,203 @@
...
+
+.. code-block:: python
+
+    >>> for ch in f.channels
+    ...     _ = ch
+
+See :ref:`Logical files` for a full list of all object types.
+
+:py:func:`dlisio.dlis.object` let you access a specific object:
```
achaikou: let(s)

python/docs/examples.rst
```diff
@@ -0,0 +1,203 @@
...
+
+Frames and Channels
+-------------------
+
+See :ref:`Curves` for information about the relationship between Channels and
+Frames. Have a look at :ref:`Channel` and :ref:`Frame`, they contain some
+useful metadata in addition to the curve-values! 
+
+Channels belonging to a Frame can be accessed directly through 
+:py:func:`dlisio.plumbing.Frame.channels`:
```
achaikou: py:attr

python/docs/examples.rst
```diff
@@ -0,0 +1,203 @@
...
+    array([852606., 852606., 852606., 852606., 852606.], dtype=float32)
+
+Access all curves in a frame with :py:func:`dlisio.plumbing.Frame.curves()`.
+The returned structured numpy array can be indexed by Channel mnemonics
+and/or sliced by samples: 
+
+.. code-block:: python
+    
+    >>> curves = fr.curves()
+    >>> curves[['TDEP', 'TENS_SL']][0:5]
```
achaikou: [ [ ] ] [ ] structure for this example may be too confusing. I from the first glance wouldn't be able to tell you what [ [ ] ] gives me

python/docs/index.rst
```diff
@@ -1,18 +1,34 @@
...
-   contain the root `toctree` directive.
-
 Welcome to dlisio's documentation!
 ==================================
 
-Contents:
+For help, examples and reference, type ``help(function)`` in
+your favourite python interpreter, or ``pydoc function`` in the unix console.
+
+Before you get started we recommended that you familiarize ourself with some
```
achaikou: you familiarize ourself - this is something new :)

python/docs/index.rst
```diff
@@ -1,18 +1,34 @@
...
+basic concepts of the dlis file format. Remember, dlis is a none-trivial format
+to work with.  Even though dlisio aim to be simple-to-use library that
```
achaikou: dlisio aim(s)

python/docs/index.rst
```diff
@@ -1,18 +1,34 @@
...
+abstracts away some of the pain of dlis, there are some concepts from the
+standard you need to be familiar with to be able work with these files.
```
achaikou: to be able (to) work

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
+About the project
+=================
+
+Welcome to dlisio. dlisio is a python package for reading Digital Log
+Interchange Standard (DLIS) v1. A version 2 exist, and has been around for
```
achaikou: (A?) version 2 exist(s)

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+quite a while, but it is our understanding that most dlis files out there are
+still version 1. Hence dlisio's focus is put on version 1 [1]_, for now.
+
+dlisio is an attempt at abstract away a lot of the pain of dlis and give access
```
achaikou: I think 'attempt to abstract away ... and give' or 'attempt in abstracting away ... and giving'

Btw, my English is not the best, so it might be good if these header files are proofread by someone else.

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+to the data in a simple and easy-to-use manner. It gives the user the ability
+to work with dlis-files without having to know the details of the standard
+itself. Its main focus is making the data accessible while putting little
+assumptions on how the data is to be used. 
+
+dlisio is written and maintained by Equinor ASA as a free, simple, easy-to-use
+way of reading well logs that can be tailored to our needs, and as a
```
achaikou: What is the meaning and purpose of "can be tailored to our needs"?

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+dlis is a binary file format for well logs, developed by Schlumberger in the
+late 80's and published by the American Petroleum Institute (API) in 1991. It
+is now the Petroleum Open Standards Consortium (POSC) [2]_ that has the
+stewardship. 
+
+When developing dlis standard, the main emphasis was put on easy writing. The
+files are structured in such a way that files can be written directly while
+acquiring the logs. This is very handy for the producers of the files, and
+equally tedious for the consumer that want to read them later on.
+Additionally is a very tolerant standard. It specifies general data-structures
```
achaikou: Additionally (it) is

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+within the file, such as channels and frames, but allow the producers to modify
+these to fit their needs. It also allow for completely new structures, not
```
achaikou: It also allow(s)

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+defined by the standard itself, such as vendor-specific object-types. This all
+sums up to a fairly complex standard with a lot of quirks and weirdness to it.
+It is safe to say that dlis a particular difficult format to work with.
```
achaikou: particular(ly)?

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+
+When developing dlis standard, the main emphasis was put on easy writing. The
+files are structured in such a way that files can be written directly while
+acquiring the logs. This is very handy for the producers of the files, and
+equally tedious for the consumer that want to read them later on.
+Additionally is a very tolerant standard. It specifies general data-structures
+within the file, such as channels and frames, but allow the producers to modify
+these to fit their needs. It also allow for completely new structures, not
+defined by the standard itself, such as vendor-specific object-types. This all
+sums up to a fairly complex standard with a lot of quirks and weirdness to it.
```
achaikou: I hope no-one will get offended by this statement :)

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+It is safe to say that dlis a particular difficult format to work with.
+
+Physical and logical files
+--------------------------
+
+The dlis standard distinguish between physical files and logical files. A
```
achaikou: distinguish(es)

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+physical file would be a .dlis file. A logical file on the other hand, is a
```
achaikou: A logical file(,) on the other hand,

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+multiple logical files, where each logical file is completely independent of
+the other logical files:: 
+    
+                            physical file (.dlis)
+            --------------------------------------------------------
+           | logical file 0 | logical file 1 | ... | logical file n |
+            --------------------------------------------------------
+
+dlisio's main entrypoint, :py:func:`dlisio.load()` takes a physical file and
+returns a tuple-like object with all the logical files.
```
achaikou: Btw, not sure if we need to mention it here or in some other place, but it's worth mentioning that we do not yet support all the divisions/connections between physical and logical files, storage sets, file sets and all that other beautiful stuff.

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+
+Logical files can be thought of as a pool of metadata and curve-values that all
+are related in some way. Some common metadata objects are Fileheader, Origins,
```
achaikou: Origin()

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+Tool, Parameter and Calibration. For a full list of all metadata objects. See
```
achaikou: Frame and Channel?

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+all object-type :ref:`Object types`::
+
+                            logical file         
+            ------------------------------------------------------
+           | Fileheader | Origin | Frame1 | Channel | Frame1 | .. |
```
achaikou: This looks weird to me because of the Frame - do you mean Frame EFLR or Fdata? And why both are Frame1?
I am not sure you can make it user-friendly
I would suggest something like Fileheader - Origin - [meta]Channel - [meta]Frame - FrameData1 - FrameData2

python/docs/intro.rst
```diff
@@ -0,0 +1,93 @@
...
+------
+
+The primary data of a dlis-file are curves (or channels). A curve is the
+recorded measurements from an instrument indexed against an axis, typically
+depth or time. In dlisio, curves are accessed through Frames or
+Channel-objects. 
+
+Channel objects contain both the metadata about the curve, and the actual curve
+data. A Frame is a gathering of multiple Channels, typically acquired at the
+same run.  Both Frame- and Channel-objects implements a :code:`curves()`-method
```
achaikou: implement()

python/docs/intro.rst
ErlendHaa: The difference between FRAME EFLR and FDATA is abstracted away by dlisio and frankly the end user does not need to know about this. Its stated below that this is a simplified illustration of a logical record. It is there to illustrate to the user how think about logical files when working with dlisio, not how LF's are implemented in the standard

python/dlisio/__init__.py
```diff
@@ -13,6 +13,44 @@
...
+    A logical file is a collection of objects - in lack of a better word,
+    that are logically connected in some way. Think of a logical file as a pool
+    of objects. Some examples of objects are Channel, Frame, Fileheader, Tool.
+    They are all accessible through their own attribute.
+
+    object() and match() let you access specific objects or search for objects
+    matching a regular expression. Unknown objects such as vendor-specific
+    objects are accessible through the unknown-attribute.
+
+    Uniquely identifying an object within a logical file can be a bit
```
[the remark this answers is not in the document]
ErlendHaa: I disagree! It is from this class the user will access objects, hence it makes sense to tell them about the requirements here. Let's say someone does something like this:

```
with dlisio.load(file) as (f, *tail):
    help(f)
 ```
Now `help(f)` clearly states what object you have, and the natural way forward. Which of course is to access some objects. An to do so, this information is highly relevant i would say.

File: python/dlisio/__init__.py [hunk 25]
```diff
@@ -13,6 +13,44 @@
...
+    problematic : list
+        Duplicated objects. If dlisio is not able to uniquely identify an object
+        based on the standards definition, the object is flagged as
+        problematic.
+
+    indexedobject : dict
+        A full inventory of _all_ objects in the logical file, indexed by type.
+        Note that there are more handy ways of accessing objects than through
+        this dictionary.
+    """
```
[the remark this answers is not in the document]
ErlendHaa: The issue with this, I think, is that attributes are defined twice. Sphinx is not smart enough (as far as I could figure out) to copy the actual docstring into this attribute list. So we end up with two copies, one here (without description), and the full docs further "down". 

And even if we could get it to sync for sphinx, we'd still have duplicates in `pydoc` and `help`. Also we would need to maintain attributes two places.

python/docs/conf.py
```diff
@@ -33,6 +33,7 @@
     'sphinx.ext.doctest',
     'sphinx.ext.intersphinx',
     'sphinx.ext.imgmath',
+    'sphinx.ext.autosectionlabel'
```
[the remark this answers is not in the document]
ErlendHaa: Just experienced the same here! Need to update it in setup.py as well. But that totally unrelated so I will make another PR for it

python/docs/intro.rst
ErlendHaa: There is no hiding the fact that dlis is _highly_ non-trivial format. Many of these things we find weird, unnecessary and annoying today, probably made a lot more sense back when it was written. 

But we need to be clear about the fact that dlis is a hard format to work with, simply because we are not able to abstract away all the quirks for our end users. So they have to deal with it as-well. To some degree at least

python/docs/examples.rst
```diff
@@ -0,0 +1,205 @@
+A quick guide
+=============
+
+This is a quick guide to get you started with dlisio. Note that all classes,
```
achaikou: all classes (and) functions
sounds a bit nicer

PR #173, opened 2019-10-09
Add dlis_packflen function

jokva:
When working with records with variable-length data types, it's
necessary to first figure out how many bytes both the source and
destination is when packing.

The packflen function is analoguous to strlen, and solves this exact
problem.

In general, it will always be necsesary to either over-allocate, or scan
the input twice. This can be avoided if it's known in advance that the
input or output is fixed-sized, but in general packflen should always
work.

Some refactoring was necessary, since figuring out the size is virtually
the same thing as reading and transforming.

The packf tests are augmented to check the packflen.


python/dlisio/ext/core.cpp
```diff
@@ -371,26 +371,28 @@ noexcept (false) {
...
-                throw std::runtime_error(msg);
-            }
+            auto assert_overflow = [end](const char* ptr, int skip) {
+                if (ptr + skip > end) {
+                    const auto msg = "corrupted record: fmtstr would read past end";
+                    throw std::runtime_error(msg);
+                }
+            };
 
-            ptr += pre_size.src;
```
achaikou: It seems to me like you don't need pre_size and post_size variables anymore.
If I understand correctly, even if there are varsize data in pre-fmt or post-fmt, it shouldn't be a problem anymore.

We don't have tests dedicated for this, but you can add the following to test_fdata:
```

def test_channel_curves():
    fpath = 'data/chap4-7/iflr/all-reprcodes.dlis'
    with dlisio.load(fpath) as (f, *_):
        channel = f.object('CHANNEL', 'CH26', 10, 0)
        curves = channel.curves()
        np.testing.assert_array_equal(
            curves[0], np.array(True))
```

There are some ascii etc data before and after channel

python/dlisio/ext/core.cpp
```diff
@@ -371,26 +371,28 @@ noexcept (false) {
...
+            int src_skip, dst_skip;
```
achaikou: Note: 'skip' sounds a bit weird as you use the same variable to actually 'skip' data and read data. Naming got me quite confused.

python/dlisio/ext/core.cpp
```diff
@@ -339,9 +352,12 @@ noexcept (false) {
...
 
-    auto pre_size  = fdata_size(pre_fmt);
-    auto data_size = fdata_size(fmt);
-    auto post_size = fdata_size(post_fmt);
+    check_supported_fmtstr(pre_fmt);
+    check_supported_fmtstr(fmt);
+    check_supported_fmtstr(post_fmt);
+
+    const auto fmt_str = std::string(fmt);
+    const auto fmt_msg = "invalid format specifier in " + fmt_str;
```
achaikou: Unused

python/dlisio/ext/core.cpp
```diff
@@ -295,33 +295,46 @@ std::string fingerprint(const std::string& type,
...
-        throw std::invalid_argument(fmt_msg);
-    }
-
-    if (variable) {
-        throw dl::not_implemented(fmt_msg);
-    }
-
-    int src = 0;
-    int dst = 0;
-    err = dlis_pack_size(fmt, &src, &dst);
```
achaikou: Can (and should) dlis_pack_size and dlis_pack_varsize be removed now?

lib/src/dlisio.cpp
```diff
@@ -556,42 +557,62 @@ namespace {
...
 }
 
 template < typename T, typename... Ts >
-char* pack( char* dst, const T* ptr, const Ts* ... ptrs ) noexcept (true) {
-    std::memcpy( dst, ptr, sizeof( T ) );
-    dst += sizeof( T );
+int pack(char* dst, const T* ptr, const Ts* ... ptrs) noexcept (true) {
+    constexpr int size = sizeof(T);
+    if (dst) {
+        std::memcpy(dst, ptr, sizeof(T));
```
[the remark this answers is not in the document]
jokva: Yes, and that was always the intention.

lib/src/dlisio.cpp
```diff
@@ -647,64 +669,87 @@ noexcept (true)
...
 }
 
-}
-
-int dlis_packf( const char* fmt, const void* src, void* dst ) {
-    cursor cur = {
-        static_cast< const char* >( src ),
-        static_cast< char* >( dst ),
-    };
+cursor dlis_packf_dispatch(const char* fmt, const char* src, char* dst) noexcept (true) {
```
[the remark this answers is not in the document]
jokva: Public functions are prefixed to provide a C idea of namespacing. This is not exposed, you're right, but it's the implementation helper _to_ a public function.

I don't mind removing the dlis_ prefix at all.

python/dlisio/ext/core.cpp
jokva: We don't _right now_, but in the very near future, fixed-size records will get a different code path, and they will come back.

python/dlisio/ext/core.cpp
jokva: I don't think so, they're still quite useful. While not useful when actually parsing records, they're useful for interrogating a few key properties from format strings. They can also be used when figuring out exactly what code path to take.

ErlendHaa:
closes #164

PR #180, opened 2019-10-21
Fdata tests fixups

achaikou:


python/tests/test_fdata.py
```diff
@@ -492,3 +507,29 @@ def test_fdata_dimension(tmpdir_factory, merge_files_manyLR):
...
+
+@pytest.mark.xfail(strict=True)
+def test_fdata_dimensions_in_multifdata():
+    fpath = 'data/chap4-7/iflr/multidimensions-multifdata.dlis'
+
+    with dlisio.load(fpath) as (f, *_):
+        frame = f.object('FRAME', 'FRAME-DIMENSION', 11, 0)
+        curves = frame.curves()
+
+        assert list(curves[0][0][0]) == [1, 2, 3]
```
ErlendHaa: Use numpy.testing.assert_array_equal [1], it produces a way cleaner and more informative output when failing

[1] https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_array_equal.html

python/data/chap4-7/README.rst
```diff
@@ -78,12 +76,26 @@ Filename                                         Description
 ================================================ ==================================================
 all-reprcodes.dlis                               1 Frame containing 27 Channels, each with a
                                                  different representation code
+                                                 
+big-ascii.dlis                                   Ascii with around 2000 characters
+
+broken-ascii.dlis                                Ascii with incorrect specified length. Around 240
```
ErlendHaa: ..., but only 20 is actually present

PR #200, opened 2019-11-28
Parse objects on demand

ErlendHaa:


python/dlisio/ext/core.cpp
```diff
@@ -835,6 +835,21 @@ PYBIND11_MODULE(core, m) {
         })
     ;
 
+    m.def( "parse_set_types", [](const std::vector< dl::record >& recs ) {
```
achaikou: Docs might be useful as it's not clear that it would return
'FILE-HEADER', 'CHANNEL', 'CHANNEL', 'encrypted', 'FRAME' 
sequential, but with omitted iflrs.

python/dlisio/plumbing/basicobject.py
```diff
@@ -54,82 +54,29 @@ class BasicObject():
...
         self.name       = name
         self.origin     = None
         self.copynumber = None
+
+        if obj is None: obj = {}
         self.attic      = obj
 
+        self.logicalfile = lf
+
         #: Dictionary with all unexpected attributes.
```
achaikou: You probably want to remove this doc

python/tests/test_dlis.py
```diff
@@ -185,3 +131,50 @@ def test_indexedobjects(f):
...
+    with dlisio.load(fpath) as (f, *tail):
+        fp = core.fingerprint('PARAMETER', 'PARAM1', 10, 0)
+        parameters = f.parameters
+
+        assert len(parameters) == 3
+        assert fp in f.indexedobjects['PARAMETER']
+
+def test_loading_new_objecttype(fpath):
+    with dlisio.load(fpath) as (f, *tail):
+        fp = core.fingerprint('TOOL', 'TOOL1', 10, 0)
```
achaikou: Depending on the test intention it might be good to check first that `fp not in f.indexedobjects['TOOL']`
plus: you will make sure that this particular action will load the object
minus: maybe it shouldn't matter when the object is loaded as long as it's loaded when needed

python/dlisio/plumbing/basicobject.py
```diff
@@ -187,16 +188,18 @@ def __repr__(self):
...
+        Intended to be called internally from the objects, properties. For
+        end-users the preferred way of reaching object-attributes are through their
         properties.
 
         Parse attributes from attic based on parsing rules defined in
-        :attr:`attributes`. If attribute contains a reference to another object that objects its
-        looked for in the parent files object-pool.
+        :attr:`attributes`. Attributes containing references to other objects
+        use :attr:`linkage` to resolve the references and create fingerprints
+        which are the used to look up the objects in the :attr:`logicalfile`'s
```
achaikou: the(n)

python/tests/test_tweaking.py
```diff
@@ -135,11 +136,8 @@ def test_dynamic_class_attribute():
     ch2.attic['PROPERTIES'] = [10]
 
     # The change should be observable on all newly created objects
-    try:
-        assert ch2.properties == 10
-        assert ch2.properties == 10
-    finally:
-        del Channel.attributes['PROPERTIES']
+    assert ch2.properties == 10
+    assert ch2.properties == 10
```
achaikou: And now you never revert channel.Properties

python/tests/test_tweaking.py
```diff
@@ -90,95 +109,79 @@ def test_remove_object_type(f):
...
+        # check that other object of the same type is not affected
+        assert ch2.source == tool
+    finally:
+        ch1.linkage = Channel.linkage
 
-        c.link(objects)
+def test_dynamic_class_linkage(f, assert_log):
+    # Change how SOURCE is parsed for all Channels
+    original = dict(Parameter.linkage)
+    Parameter.linkage['AXIS'] = linkage.objref
```
achaikou: Would be nice to move this also in the try :)

PR #220, opened 2019-12-17
Return mnemonic not Channel for frame/channel.index

### LINKED ISSUE #199: Fallback to FRAMENO in frame.index and channel.index if no index channel
From to 5.7.1 of rp66v1 [1] :

> If Attribute Index-Type is absent, then there is no Index Channel and Attributes Direction and Spacing are meaningless and are ignored. When Attribute Index-Type is absent, then Frames are implicitly indexed by the Frame Number. 

This logic is implemented but None is returned instead of FRAMENO's in the no-index case. With #185 returning FRAMENO is trivial.

[1] http://w3.energistics.org/RP66/V1/Toc/main.html

ErlendHaa:
With FRAMENO integrated into the numpy.ndarray returned by .curves()
it's natural to return FRAMENO, not None, from .index if there is no
index Channel. However, FRAMENO is not a Channel object. To make the
index properties coherent with FRAMENO, Channel.index and Frame.index
now returns the channel mnemonic, rather than the full channel object.

This is a breaking change, and some workflows where the user works
directly with Channels and not through the Frame objects will be a bit
more cumbersome. However, index now works seamlessly with indexing the
.curves() array, e.g:

``` 
curves = frame.curves()
 curves[[frame.index, 'GR', 'ALH1']]
```

closes #199

### DIFF
--- a/python/dlisio/plumbing/channel.py
+++ b/python/dlisio/plumbing/channel.py
@@ -129,22 +129,6 @@ class Channel(BasicObject):
         return self['SOURCE']
 
     @property
-    def index(self):
-        """The Channel this Channel is indexed against, if any. See
-        :attr:`Frame.index_type` for definition of existing index channel.
-
-        Returns
-        -------
-
-        channel : Channel or None
-        """
-        msg = 'This channel is an index channel'
-        index = self.frame.index
-
-        if index == self: logging.info(msg)
-        return index
-
-    @property
     def dtype(self):
         """dtype
 
--- a/python/dlisio/plumbing/frame.py
+++ b/python/dlisio/plumbing/frame.py
@@ -143,28 +143,23 @@ class Frame(BasicObject):
 
     @property
     def index(self):
-        """The Channel that serves as an index for all other Channels in this
-        Frame.
-
-        Frames may or may not have an index Channel, see :attr:`index_type` for
-        definition of existing index channel.
+        """Mnemonic of the channel all channels in this Frame are indexed
+        against, if any. See :attr:`Frame.index_type` for definition of
+        existing index channel.
 
         Returns
         -------
-        channel : Channel or None
-        """
-        msg = 'There is no index channel, see help(frame.index) for more info'
-        index = None
 
-        if self.index_type is None:
-            logging.warning(msg)
+        mnemonic : str
+        """
+        msg = 'Frame has no channels'
+        index = None
+        if len(self.channels) == 0:
+            logging.info(msg)
             return index
 
-        try:
-            index = self.channels[0]
-        except IndexError:
-            logging.warning(msg)
-
+        if self.index_type is None: index = 'FRAMENO'
+        else:                       index = self.channels[0].name
         return index
 
     @property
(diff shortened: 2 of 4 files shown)


python/dlisio/plumbing/channel.py
```diff
@@ -130,18 +130,18 @@ def source(self):
 
     @property
     def index(self):
```
achaikou: Btw, I agree with Jørgen that this property shouldn't be present for channels and should be accessed through frame.

python/dlisio/plumbing/frame.py
```diff
@@ -143,28 +143,22 @@ def index_max(self):
 
     @property
     def index(self):
-        """The Channel that serves as an index for all other Channels in this
-        Frame.
-
-        Frames may or may not have an index Channel, see :attr:`index_type` for
-        definition of existing index channel.
+        """mnemonic of the channel this Channel is indexed against, if any. See
```
achaikou: this (Frame) is indexed against

And maybe start with uppercase (M)nemonic?

python/dlisio/plumbing/channel.py
```diff
@@ -130,18 +130,18 @@ def source(self):
...
-        """The Channel this Channel is indexed against, if any. See
+        """mnemonic of the channel this Channel is indexed against, if any. See
```
achaikou: (M)nemonic
 And probably remove "if any". Now there is always something

python/dlisio/plumbing/frame.py
```diff
@@ -396,19 +390,8 @@ def curves(self):
...
-        >>>     pdcurves.index.name = indexname
-        >>> else:
-        >>>     pdcurves = pd.DataFrame(curves)
-
-        Let's walk through this. Firstly, check that *index_type* is not None,
-        if it is, there is no index channel to set. Secondly, note how the
-        curvenames are all stored in the array itself. Then all that is needed
-        is to extract the first name as index and let the rest be added to the
-        DataFrame as normal Channels.
+        >>> pdcurves = pd.DataFrame(curves[curves], index=curves[f.index])
```
achaikou: Do I understand correctly that index channel (or frameno) will appear twice? As index as as part of the data?

python/dlisio/plumbing/channel.py
ErlendHaa: If any, as in "if there is a index _Channel_"

python/dlisio/plumbing/frame.py
```diff
@@ -143,28 +143,23 @@ def index_max(self):
 
     @property
     def index(self):
-        """The Channel that serves as an index for all other Channels in this
-        Frame.
-
-        Frames may or may not have an index Channel, see :attr:`index_type` for
-        definition of existing index channel.
+        """Mnemonic of the channel all channels in this Frame is indexed
```
achaikou: Frame (are) undexed

python/dlisio/plumbing/frame.py
```diff
@@ -396,19 +391,8 @@ def curves(self):
...
-        >>>     pdcurves.index.name = indexname
-        >>> else:
-        >>>     pdcurves = pd.DataFrame(curves)
-
-        Let's walk through this. Firstly, check that *index_type* is not None,
-        if it is, there is no index channel to set. Secondly, note how the
-        curvenames are all stored in the array itself. Then all that is needed
-        is to extract the first name as index and let the rest be added to the
-        DataFrame as normal Channels.
+        >>> pdcurves = pd.DataFrame(curves.dtype.names[1:], index=curves[f.index])
```
achaikou: Shouldn't it be
`curves[list(curves.dtype.names[1:])] `
then? I am getting lost 

Maybe repeated index is easier for the user?

ErlendHaa: I'd go for the repeated index for simplicity
(And yes, it should have been a list)

PR #229, opened 2020-01-06
Remove cache of dtype & fmtstr in Channel & Frame

### LINKED ISSUE #203: Consider removing cache from object
We often have lines like that as the first line of the method:
`if self._something: return self._something`

In many cases we support patching, like adding your own classes or changing existing values.

But caching (with no direct way of removing this cache) basically says: whatever was counted once will never be recounted again. So when user does sequence of actions:

1. get the cached property
2. do some patching affecting what property should be
3. get property again

User with receive old cached property anyway.

So we either should get rid of caching or explicitly state that all the patching is supported only before any actual work begins.

ErlendHaa:
Caching these variables us little value. If anything its a source for
unexpected behavior (e.g. if someone would manually change them). The
computational overhead of having to re-compute these variable is small
enough that removing the cache should not have a significant impact on
performance.

closes #203

### DIFF
--- a/python/dlisio/plumbing/channel.py
+++ b/python/dlisio/plumbing/channel.py
@@ -87,9 +87,6 @@ class Channel(BasicObject):
     def __init__(self, obj = None, name = None, lf=None):
         super().__init__(obj, name = name, type = 'CHANNEL', lf=lf)
         # The numpy data type of the sample array
-        self._dtype        = None
-        # Format-string of the channel. Mainly intended for internal use
-        self._fmtstr       = None
         self._frame        = None
 
     @property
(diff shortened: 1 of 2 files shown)


python/dlisio/plumbing/channel.py
```diff
@@ -140,14 +137,12 @@ def dtype(self):
 
         dtype : np.dtype
         """
-        if self._dtype: return self._dtype
-
         if self.dimension == [1]:
-            self._dtype = np.dtype(dtype[self.reprc])
+            ch_dtype = np.dtype(dtype[self.reprc])
```
jokva: Can't we just return directly here?

Issue #233, opened 2020-01-10
dlis.object() erroneously reports multiple parameters with the same name


erlendviggen:
I've found the dlis.object() function to be a very convenient way to get channels with unique names in a file. However, when I try to do this with parameters, the function throws an error:

    file.object('parameter', 'FN')

results in the error 

    ValueError: There are multiple parameters named FN. Found: (origin=57, copy=0), (origin=57, copy=0), 

I get the same error for any parameter I try. The parameters are not in fact multiply defined; this code returns `1` as you would expect:

    parameters = [par for par in file.parameters if par.name == 'FN']
    len(parameters)

I have verified that the error occurs on multiple log files from the public [Volve Data Village](https://www.equinor.com/en/how-and-why/digitalisation-in-our-dna/volve-field-data-village-download.html) data set. In no case did the error not occur.

This is of course not a serious problem, as the second approach works quite well for picking out named parameters as well. But is this problem due to a bug in dlisio, or am I simply using dlis.object() wrong?

jokva:
This looks like a bug.

ErlendHaa:
Definitely a bug, fixed it in #234, which is a part of v0.1.16 [1] that just got released

[1] https://github.com/equinor/dlisio/blob/master/CHANGELOG.md

erlendviggen:
I can confirm that this now works perfectly for me in 0.1.16.

PR #234, opened 2020-01-14
Fix a bug where match() yielded the same obj more than once

ErlendHaa:
The dlis class keeps track of the type of object_sets that are present
in the file. This list is being used by dlis.match to match on
object-type. Because this is a list, and not a set, match would search a
particular object-type multiple times if there where multiple
object_sets of that type.

`dlis.object()` was effected in the same way as it is powered by `dlis.match()`


ErlendHaa:
Closes #233

Issue #230, opened 2020-01-08
Consider adding the number of parameters to dlis.describe()


erlendviggen:
The `describe()` function of `dlis` objects currently prints the number of frames and channels in the logical file, but not the number of parameters. It's quite easy for the user to find this number as `len(file.parameters)`, but maybe it should also be printed in the overview? It's a tiny thing, but I thought I might mention it.

ErlendHaa:
Hi

This feature got removed in 0.1.15 actually. It kind of a side-effect of an performance-improvement we introduced. Let me explain:

Pre 0.1.15, `dlisio.load()` parsed and loaded _every_ metadata object in the file. Doing so was one of the bottlenecks of load, especially for files with a high number of metadata objects. In 0.1.15, only Fileheader, Origin, Frame and Channel-objects are loaded by default. Other object-types are loaded when need, e.g. `len(f.parameter)` will load parameter object. 

Due to the way objects are structured in dlis-files, it is as expensive to just count them as it would be to parse them (i.e. do a full load). Hence if `dlis.describe()` with object-count for all objects would eat up all the performance gained in `dlisio.load`. 

A option might be to let `dlis.describe()` take an argument that specifies whether or not to count all objects.

ErlendHaa:
On a second thought, if someone call `.describe()` it is reasonable to assume that they are not processing a large amount of dlis-files, but rather focuses on exploring a single (or a few) files. Hence, the performance gain is not that important anymore and we should probably prioritize convenience.

ErlendHaa:
Brought this back in #235, its part of v0.1.16

PR #235, opened 2020-01-14
Reintroduce object-counts in dlis.describe()

ErlendHaa:
Due to the just-in-time loading of objects introduced in [1], the
object-count for each object-type in dlis.describe() was removed as it
would in effect load every object in the file.

Seaming that this information is valuable it is re-introduced on the
grounds that its fair to assume that describe() is only called in the
context of exploring a single file, or a few files. Hence, we can
tolerate the performance hit.

[1] SHA: f5090cc67b1d12add8ca4dab44bd0278bcde01f9

### DIFF
--- a/python/dlisio/__init__.py
+++ b/python/dlisio/__init__.py
@@ -545,22 +545,20 @@ class dlis(object):
 
         plumbing.describe_dict(buf, d, width, indent)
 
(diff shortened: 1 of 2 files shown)


ErlendHaa:
closes #230

PR #238, opened 2020-01-16
Allow duplicated mnemonics in curves()

ShikhaMishra9:
By default we allow duplicated mnemonics as long as Channels are
distinguishable by the triplet; mnemonic, origin, copynumber. If that is
not the case, its a clear standard violation and curves() throws.

Apparently there are files where this is the case, hence we supply the
option to read the curve-data anyway. The default-behavior of curves()
is still to throw if the standard is violated.

### DIFF
--- a/python/dlisio/plumbing/frame.py
+++ b/python/dlisio/plumbing/frame.py
@@ -155,8 +155,7 @@ class Frame(BasicObject):
         else:                       index = self.channels[0].name
         return index
 
-    @property
-    def dtype(self):
+    def dtype(self, strict=True):
         """dtype
 
         data-type of each frame, i.e. the sum of channel.dtype of each channel
@@ -252,7 +251,10 @@ class Frame(BasicObject):
             dtype = np.dtype(types)
         except ValueError as exc:
             logging.error(duplerr.format(self.name, exc))
-            raise
+            if strict: raise
+
+            types = mkunique(types)
+            dtype = np.dtype(types)
 
         return dtype
 
@@ -273,17 +275,35 @@ class Frame(BasicObject):
         # variable-lenght unsigned integer (i).
         return 'i' + ''.join([x.fmtstr() for x in self.channels])
 
-    def curves(self):
+    def curves(self, strict=True):
         """All curves belonging to this frame
 
         Get all the curves in this frame as a structured numpy array. The frame
         includes the frame number (FRAMENO), to detect errors such as missing
         entries and out-of-order frames.
 
+        Parameters
+        ----------
+
+        strict : boolean, optional
+            By default (strict=True) curves() raises a ValueError if there are
+            multiple channel with the same values for both name, origin and
+            copynumber. This would be a clear violation of the dlis-spec.
+            Setting strict=False lifts this restriction and dlisio will append
+            numerical values (i.e. 0, 1, 2 ..) to the labels used for
+            column-names in the returned array.
+
         Returns
         -------
         curves : np.ndarray
             curves with dtype = self.dtype
+
+        Raises
+        ------
+
+        ValueError
+            If there multiple channels with identical name, origin, copynumber
+            in Frame.channels. This can be suppressed by passing strict=False
 
         See also
         --------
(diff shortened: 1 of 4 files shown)


python/dlisio/plumbing/frame.py
```diff
@@ -379,8 +399,52 @@ def curves(self):
...
+        Sometimes a Frame can contain multiple Channels with the same
+        name/mnemonic, in that case the labels are augmented with the origin and
+        copynumber:
+
+         >>> frame.channels
+        [Channel(TDEP), Channel(TDEP), Channel(GR)]
+
+        >>> curves = frame.curves()
+        >>> curves.dtype.names
+        ('FRAMENO', 'TDEP.0.0', 'TIME.0.1', 'GR')
```
ErlendHaa: `('FRAMENO', 'TDEP.0.0', 'TDEP.0.1', 'GR')`

python/dlisio/plumbing/frame.py
```diff
@@ -379,8 +399,52 @@ def curves(self):
...
+
+        >>> Frame.curves()
+        ValueError: field 'TDEP.0.0' occurs more than once
+
+        You can still get the underlying data by passing strict=False, then
+        dlisio will append numerical values to the duplicated names like so:
+
+        >>> curves = frame.curves(strict=False)
+        >>> curves.dtype.names
+        ('FRAMENO', 'TDEP.0.0(0)', 'TIME.0.1(1)', 'GR')
```
ErlendHaa: `('FRAMENO', 'TDEP.0.0(0)', 'TDEP.0.0(1)', 'GR')`

python/dlisio/plumbing/frame.py
```diff
@@ -273,18 +275,36 @@ def fmtstr(self):
...
         includes the frame number (FRAMENO), to detect errors such as missing
         entries and out-of-order frames.
 
+        Parameters
+        ----------
+
+        strict : boolean, optional
+            By default (strict=True) curves() raises a ValueError if there are
+            multiple channel with the same values for both name, origin and
+            copynumber. This is a clear violation. Setting strict=False lifts
```
ErlendHaa: This would be a clear violation of the dlis-spec

python/dlisio/plumbing/frame.py
```diff
@@ -379,8 +399,52 @@ def curves(self):
         >>> curves = frame.curves()
         >>> pdcurves = pd.DataFrame(curves, index=curves[f.index])
         >>> pdcurves.index.name = f.index
+
+        By default the returned np.ndarray have column-names that reflects the
+        mnemonics of each Channel
+
+        >>> frame.channels
+        [Channel(TDEP), Channel(TDEP), Channel(GR)]
```
ErlendHaa: `[Channel(TDEP), Channel(TIME), Channel(GR)]`

python/dlisio/plumbing/frame.py
```diff
@@ -379,8 +399,52 @@ def curves(self):
...
+
+        >>> curves = frame.curves()
+        >>> curves.dtype.names
+        ('FRAMENO', 'TDEP', 'TDEP', 'GR')
```
ErlendHaa: `('FRAMENO', 'TDEP', 'TIME', 'GR')`

python/dlisio/plumbing/frame.py
```diff
@@ -379,8 +399,52 @@ def curves(self):
...
+
+        >>> curves = frame.curves()
+        >>> curves.dtype.names
+        ('FRAMENO', 'TDEP.0.0', 'TDEP.0.1', 'GR')
+
+        This augmentation is customizable by changing :attr:`dtype_fmt`. See
+        :attr:`dtype`. However, in some cases even this is not enough to make
+        unique labels for the columns. That would be a clear violation of the
+        dlis spec and :attr:`curves` will raise an ValueError:
+
```
ErlendHaa: This augmentation is customizable by changing :attr:`dtype_fmt`. See :attr:`dtype`. Some frames have multiple instance of the same channel or mulitple channels where name, origin and copynumber are identical. This is a clear violation of the dlis spec and :attr:`curves()` will raise an ValueError by default.

python/dlisio/plumbing/frame.py
```diff
@@ -379,8 +399,52 @@ def curves(self):
...
+        >>> frame.channels
+        [Channel(TDEP), Channel(TDEP), Channel(GR)]
+
+        >>> Frame.curves()
+        ValueError: field 'TDEP.0.0' occurs more than once
+
+        You can still get the underlying data by passing strict=False, then
```
ErlendHaa: However, :func:`curves` offers an escape-hatch to get the underlying data. By passing strict=False dlisio appends numerical values to the identical channels

PR #248, opened 2020-03-17
changing exception to warning for diff repcode

ShikhaMishra9:
Throwing a warning and supplying default value based on the
new reprcode.

### DIFF
--- a/python/tests/test_logical_record.py
+++ b/python/tests/test_logical_record.py
@@ -113,8 +113,6 @@ def test_absent_attribute_in_template(tmpdir, merge_files_oneLR):
         obj = f.object('VERY_MUCH_TESTY_SET', 'OBJECT', 1, 1)
         assert obj.attic['DEFAULT_ATTRIBUTE']
 
-
-
 @pytest.mark.future_test_attributes
 def test_global_default_attribute(tmpdir, merge_files_oneLR):
     path = os.path.join(str(tmpdir), 'global-default-attribute.dlis')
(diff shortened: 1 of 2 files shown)


python/tests/test_logical_record.py
```diff
@@ -113,7 +113,20 @@ def test_absent_attribute_in_template(tmpdir, merge_files_oneLR):
...
+        'data/chap3/start.dlis.part',
+        'data/chap3/template/default.dlis.part',
+        'data/chap3/object/object.dlis.part',
+        'data/chap3/objattr/csingle-novalue.dlis.part'
+    ]
+    merge_files_oneLR(path, content)
 
+    with dlisio.load(path) as (f, *_):
+        obj = f.object('VERY_MUCH_TESTY_SET', 'OBJECT', 1, 1)
+        assert obj.attic['DEFAULT_ATTRIBUTE']
```
achaikou: OK, there is one problem with this.
We do not report an error anymore, which means that
`obj.attic['DEFAULT_ATTRIBUTE'] `equals something. And we want this something to make sense, so we have to check it.

Right now in the code it equals [-0.75, 10.0] because this is the value in the template (float). But we changed the type from float to csingle, we shouldn't be seeing this value anymore!
in `parse.cpp/object_vector parse_objects` we use `patch_missing_value( attr.value, count, attr.reprc )` after reporting an error. Problem is that `patch_missing_value` does nothing if there already was a value set in template. It then leaves it to be an old one. In case of the test, floats.

What we want in case of such error is to disregard whatever is in the template and create a default value of new type.
You can do it by adding
`attr.value = mpark::monostate{};`
right after reporting the warning. This line says "ok, I declare that I have no value whatsoever", thus allowing us to supply it with default one in `patch_missing_value`.
Then the check in the test would be correct to the test logic
`assert obj.attic['DEFAULT_ATTRIBUTE'] == [0j, 0j]`

python/tests/test_logical_record.py
```diff
@@ -113,7 +113,20 @@ def test_absent_attribute_in_template(tmpdir, merge_files_oneLR):
         obj = f.object('VERY_MUCH_TESTY_SET', 'OBJECT', 1, 1)
         assert obj.attic['DEFAULT_ATTRIBUTE']
 
+@pytest.mark.future_warning_repcode_different_no_value
```
achaikou: :+1:

python/tests/test_logical_record.py
```diff
@@ -302,6 +282,20 @@ def test_repcode_invalid_in_objects(tmpdir, merge_files_oneLR):
...
+        'data/chap3/start.dlis.part',
+        'data/chap3/template/default.dlis.part',
+        'data/chap3/object/object.dlis.part',
+        'data/chap3/objattr/csingle-novalue.dlis.part'
+    ]
+    merge_files_oneLR(path, content)
+
+    with dlisio.load(path) as (f, *_):
+        obj = f.object('VERY_MUCH_TESTY_SET', 'OBJECT', 1, 1)
+        assert obj.attic['DEFAULT_ATTRIBUTE']
```
achaikou: This check is not enough. It just checks that DEFAULT_ATTRIBUTE value is not empty.
But we do not check what this value is. Is it still some old floats, or is it now new default csingles, as we want them? We still do not prove that monostate solution worked.

So please check that it equals to complex default values, as per previous comment.

achaikou:
Great! A small note about commit message, and then we should be done:

commit title: we usually throw exceptions, but report warnings. Throwing a warning sounds weird. So "Report warning for diff repcode" or "Warn on changed repcode" makes more sense to me.

Issue #255, opened 2020-04-24
ValueError: month must be in 1..12


FDB-PG:
I have >2000 borehole image logs that I am bulk processing into a single resistivity curve from all the available data within each channel. 

A number of these DLIS files return value errors 

ValueError: month must be in 1..12

I am not sure how I can get around this error, so any feedback will be much appreciated.

Thanks,
Fabian

channel_dict_ = ['DEPT','DEPTH','BT1L','BT1U','BT2L','BT2U','BT3L','BT3U','BT4L','BT4U','BT5L','BT5U','BT6L','BT6U','BT7L','BT7U','BT8L','BT8U']

total_start = time.time()

cmi = pd.DataFrame()
list_ = []
df = pd.DataFrame()

list_ = []
df = pd.DataFrame()
n = 0
with dl.load(r'H:\\well_name.dlis') as (f, *tail):
    channel_dict={ ch.name: ch for ch in f.channels }
    for ch in channel_dict_:
        try:
            c=channel_dict[ch]
            curve = np.around(c.curves(),4)
            print(c)
            list_.append(curve)
            try:
                df2 = pd.DataFrame(list_[n])
                i = 0
                for cl in df2.columns:
                    df2.rename(columns = {df2.columns[i]:ch+'_'+str(i)}, inplace = True)
                    i = i+1
            except:
                df2 = pd.DataFrame(np.reshape(list_[n], (len(list_[n]),-1)))
                i = 0
                for cl in df2.columns:
                    df2.rename(columns = {df2.columns[i]:ch+'_'+str(i)}, inplace = True)
                    i = i+1
            n = n+1
            df = pd.concat([df,df2], axis =1, sort = False)
        except (KeyError):
            pass

Traceback (most recent call last):

  File "<ipython-input-26-937c494e152a>", line 12, in <module>
    with dl.load(r'H:\\well_name.dlis') as (f, *tail):

  File "C:\Users\Fabian.Brandimarte\AppData\Local\Continuum\anaconda3\lib\site-packages\dlisio\__init__.py", line 787, in load
    part['records'], implicits, sul_offset=sulpos)

  File "C:\Users\Fabian.Brandimarte\AppData\Local\Continuum\anaconda3\lib\site-packages\dlisio\__init__.py", line 232, in __init__
    self.load(recs)

  File "C:\Users\Fabian.Brandimarte\AppData\Local\Continuum\anaconda3\lib\site-packages\dlisio\__init__.py", line 624, in load
[...]

PR #256, opened 2020-04-24
[WIP] Invalid datetime check

achaikou:
Submitting to CI to have a better picture of when this is failing and
when not.
For some reason test passes on my machine.


achaikou:
Investigation showed that #255 is unaffected with python 3.5 but fails with higher python versions.

PR #257, opened 2020-04-24
Missing tests

achaikou:
New tests (also failing) that might be useful for future changes


python/tests/test_physical_layout.py
```diff
@@ -102,6 +102,12 @@ def test_truncated_on_lrs():
         _ = dlisio.load('data/chap2/truncated-on-lrs.dlis')
     assert "file truncated" in str(excinfo.value)
 
+@pytest.mark.xfail(strict=True)
```
ErlendHaa, line 111: Maybe put the same reasoning as in the commit in the `reason` argument.

python/tests/test_physical_layout.py
```diff
@@ -33,6 +33,12 @@ def test_vrl_and_lrsh_mismatch():
         _ = dlisio.load('data/chap2/wrong-lrhs.dlis')
     assert "visible record/segment inconsistency" in str(excinfo.value)
 
+@pytest.mark.xfail(strict=True)
```
ErlendHaa, line 36: Same comment regarding the `reason` argument. Or turn it from a double fail to a pass? I'm not 100% sure that we should fail on inconsistent attributes if we dont have to, but thats a discussion for later

python/data/chap3/README.rst
```diff
@@ -20,6 +20,33 @@ start.dlis.part              SUL + all the VR and LRSH bytesiBytes 50 and 51
 /reprcode                    Object attribute of various representation codes
 ============================ =================================================
 
+implicit folder contains full files and it's worth to describe it in more
```
ErlendHaa, line 23: > and it's worth to describe it in more details.

Skip this? Its obvious that its worth describing as we have a full table below
(table is great btw

python/data/chap3/README.rst
```diff
@@ -20,6 +20,33 @@ start.dlis.part              SUL + all the VR and LRSH bytesiBytes 50 and 51
...
+
+====================================== ========================================
+Files in /implicit                     Description
+====================================== ========================================
+fdata-broken-obname.dlis               File is truncated in the middle of obname
+                                       in object description in fdata
+
+fdata-encrypted.dlis                   Fdata record is encrypted
+
+fdata-many-in-same-vr.dlis             Several fdata segments in same vr,
```
ErlendHaa, line 34: We probably want one with min_obname_size aswell (`idlen` is 0), if do not already have it?

python/data/chap3/README.rst
```diff
@@ -20,6 +20,33 @@ start.dlis.part              SUL + all the VR and LRSH bytesiBytes 50 and 51
 /reprcode                    Object attribute of various representation codes
 ============================ =================================================
 
+implicit folder contains full files and it's worth to describe it in more
+details.
+
+====================================== ========================================
+Files in /implicit                     Description
+====================================== ========================================
+fdata-broken-obname.dlis               File is truncated in the middle of obname
```
ErlendHaa, line 29: Confussing sentence. Maybe
> File is truncated i the obname of a fdata record

python/data/chap4-7/README.rst
```diff
@@ -140,6 +140,9 @@ Remaining files
 ================================================ ==================================================
 Filename                                         Description
 ================================================ ==================================================
+invalid-date-in-origin.dlis                      Simple file which contains wrong creation time
```
ErlendHaa, line 143: Simple file which contains ~~wrong~~ invalid creation time attribute in origin

python/tests/test_loading.py
```diff
@@ -108,3 +108,12 @@ def test_context_manager_with():
...
             _ = g.fileheader
 
+@pytest.mark.xfail(strict=False)
+def test_invalid_attribute_in_load():
+    # Error in one attribute shouldn't prevent whole file from loading
+    # This is based on common enough error in creation time property in
+    # origin.
+    # It loads just fine on python 3.5, but fails in higher versions
+    path = 'data/chap4-7/invalid-date-in-origin.dlis'
+    with dlisio.load(path) as (f, *_):
```
ErlendHaa, line 118: `with dlisio.load(path) as _:` to please codacy

PR #262, opened 2020-05-14
Make channel.curves more robust

### LINKED ISSUE #252: Fail more gracefully in channel.curves() when frame object is not found
`channel.curves` fails with `AttributeError: 'NoneType' object has no attribute 'curves'` if the channel does not belong to any frame. I.e. the `channel.frame` lookup used in `curves` yields no result.

We should catch such failed lookups and raise a more sensible error with a more helpful message

ErlendHaa:
resolves #252

### DIFF
--- a/python/docs/examples.rst
+++ b/python/docs/examples.rst
@@ -198,7 +198,8 @@ and/or sliced by samples:
 Note that double brackets are needed in order to access muliple channels at
 once.
 
-If you prefer to work with pandas over numpy, the conversion is trivial:
+As long as the frame only contains channels with scalar samples, it can be
+trivially converted to a pandas DataFrame:
 
 .. code-block:: python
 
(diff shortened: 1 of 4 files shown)


python/docs/examples.rst
```diff
@@ -198,7 +198,8 @@ and/or sliced by samples:
 Note that double brackets are needed in order to access muliple channels at
 once.
 
-If you prefer to work with pandas over numpy, the conversion is trivial:
+As long as the frame only contain channels with scalar samples, it can be
```
achaikou, line 201: only contain(s)

python/docs/examples.rst
```diff
@@ -198,7 +198,8 @@ and/or sliced by samples:
...
+trivially converted to a pandas DataFrame:
```
achaikou, line 202: Btw, if you rebase to your new changes, you'll see that line I suggested updating originally:

> If you prefer to work with pandas over numpy, the conversion is trivial:

PR #268, opened 2020-06-15
Close unused handle in dlisio.load

ErlendHaa:
There are no hints in the files to whether there are more logical files
to come. load therefore opens file handles for each logical file in a
loop until eof is hit. This last file handle is never used, nor where it
explicitly closed, causing load to leak it.

### DIFF
--- a/python/tests/test_loading.py
+++ b/python/tests/test_loading.py
@@ -9,6 +9,32 @@ import os
 
 import dlisio
 
+def test_filehandles_closed(tmpdir):
+    # Check that we don't leak open filehandles
+    #
+    # This test uses the fact that os.remove fails on windows if the file is in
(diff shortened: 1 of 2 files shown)


python/tests/test_loading.py
```diff
@@ -9,6 +9,32 @@
...
+    # use as a proxy for testing that dlisio dont leak filehandles.  From the
+    # python docs [1]:
+    #
+    #   On Windows, attempting to remove a file that is in use causes an
+    #   exception to be raised; on Unix, the directory entry is removed but the
+    #   storage allocated to the file is not made available until the original
+    #   file is no longer in use.
+    #
+    # On linux on the other hand, os.remove does not fail even if there are
+    # open filehandles, hence this test only makes sence on Windows.
```
achaikou, line 25: makes sen(s)e

PR #270, opened 2020-06-15
Close filehandle

achaikou:


python/dlisio/__init__.py
```diff
@@ -749,70 +749,73 @@ def load(path):
...
+                offset = core.findvrl(stream, hint)
+            except RuntimeError:
+                if stream.eof(): break
+                raise
+
+        return Batch(lfs)
+
+    finally:
+        # last object assigned to stream variable is always broken or of no
+        # interest to us, so we can safely close it
```
ErlendHaa, line 817: This is insufficient whenever something fails. Assume load has successfully gathered up two logical files in `lfs` then the third one fails on e.g. `findoffsets`, now only the last/current one is released, not the two in `lfs`.

This would not be an issue if we actually _returned_  the successfully indexed logical files, so maybe we address that?

achaikou: oh, you are so right. totally missed that :woman_facepalming:

python/dlisio/__init__.py
```diff
@@ -749,70 +749,76 @@ def load(path):
...
+    stream = open(path)
+    try:
+        try:
+            offset = core.findsul(stream)
+            sul = stream.get(bytearray(sulsize), offset, sulsize)
+            offset += sulsize
+        except:
+            offset = 0
+            sul = None
+
```
ErlendHaa, line 766: Would move the outer-most try down here, as the code above never raises.

python/dlisio/__init__.py
```diff
@@ -749,70 +749,76 @@ def load(path):
...
+        tapemarks = core.hastapemark(stream)
+        offset = core.findvrl(stream, offset)
+
+        def rewind(offset, tif):
```
ErlendHaa, line 770: Move out of try

PR #273, opened 2020-06-17
Cleanup

achaikou:


lib/src/io.cpp
```diff
@@ -22,7 +22,7 @@ stream open(const std::string& path, std::int64_t offset) noexcept (false) {
     auto* file = std::fopen(path.c_str(), "rb");
     auto* protocol = lfp_cfile(file);
     if ( protocol == nullptr  )
-        throw io_error("lfp: unable to open lfp protocol cfile");
+        throw io_error("lfp: unable to open file for path " + path);
```
ErlendHaa, line 25: `std::fopen` is actually the one that fails when the file does not exists:
> If successful, returns a pointer to the object that controls the opened file stream, with both eof and error bits cleared. The stream is fully buffered unless filename refers to an interactive device.
On error, returns a null pointer. POSIX requires that errno is set in this case. [1]

It just happens to be that if a nullptr is passed to `lfp_cfile` a nullptr is returned as well. So would do
```
if (!file)
   throw("unable to open file")
```
and leave the protocol error where it is. 

Also, if we want to be _really_ helpful we can switch on (some) of the error codes [2] from `fopen`. Then we can state "Unable to open file: No such file or directory" with confidence

[1] https://en.cppreference.com/w/cpp/io/c/fopen
[2] https://pubs.opengroup.org/onlinepubs/9699919799/functions/fopen.html

python/tests/test_tif.py
```diff
@@ -91,3 +91,21 @@ def test_layout_fdata_disaligned():
         frame = f.object('FRAME', 'FRAME-REPRCODE', 10, 0)
         curves = frame.curves()
         assert len(curves) == 1
+
+def test_layout_truncated_in_data():
+    fpath = 'data/tif/layout/truncated-in-data.dlis'
```
ErlendHaa, line 96: Can we put these file in the README for completeness?

ErlendHaa:
The output on travis was extremely unhelpful so ran the tests on my machine.
```
def test_load_nonexisting_file():
    with pytest.raises(OSError) as exc:
        ...
```
fails with `RuntimeError: lfp: unable to open file for path this_file_does_not_exist.dlis`. (different exception type)

PR #277, opened 2020-06-24
[not to be merged] Circleci 0a

achaikou:
CircleCI run on Windows modifies our binary files causing tests to fail. 

Didn't find this mentioned on the forum, so would try to submit it if able to beat the login to their forum :crying_cat_face: As of now I am totally defeated by 2FA.


achaikou:
Looks like we need to manually deal with git/newline options on CircleCI.
We didn't seem to do it on appveyor, probably they set these settings as default.

Issue #278, opened 2020-07-05
Infinite loop in dl::findoffsets?


ejschoen:
We're using version 0.2.3 to grab metadata from explicit frames from a client's dlis files.  We're running into issues scanning file at least one file that from its filename would appear to be a Schlumberger run with Rt Scanner and Sonic Scanner (ZAIT, MSIP), logged within the last year (so from Maxwell).   The file is approximately 500MB in size.  Calls to `dlisio.load()` never return on this file.  

Using htop's strace capability, we can see that the only system calls being made in the process are to `seek` at a particular offset (`200282112`).  Running under gdb and interrupting the process after a few seconds, the stack looks like this:

```
#0  __lseek64 (fd=3, offset=200282112, whence=0) at ../sysdeps/unix/sysv/linux/lseek64.c:36
#1  0x00007ffff7e60eff in _IO_new_file_seekoff (fp=0xa23e00, offset=200281176, dir=0, mode=<optimized out>) at libioP.h:948
#2  0x00007ffff7e5d4fd in __GI_fseek (fp=0xa23e00, offset=200281176, whence=0) at fseek.c:36
#3  0x00007ffff69e96f1 in ?? () from py/.dlis/lib/python3.8/site-packages/dlisio/../dlisio.libs/liblfp-197c740c.so
#4  0x00007ffff69f18ad in ?? () from py/.dlis/lib/python3.8/site-packages/dlisio/../dlisio.libs/liblfp-197c740c.so
#5  0x00007ffff69dc6c8 in lfp_seek () from py/.dlis/lib/python3.8/site-packages/dlisio/../dlisio.libs/liblfp-197c740c.so
#6  0x00007ffff6e651a1 in dl::stream::seek(long) () from py/.dlis/lib/python3.8/site-packages/dlisio/../dlisio.libs/libdlisio-extension-f45da5f4.so
#7  0x00007ffff6e655f2 in dl::findoffsets(dl::stream&) () from py/.dlis/lib/python3.8/site-packages/dlisio/../dlisio.libs/libdlisio-extension-f45da5f4.so
#8  0x00007ffff73602f8 in ?? () from py/.dlis/lib/python3.8/site-packages/dlisio/core.cpython-38-x86_64-linux-gnu.so
#9  0x00007ffff7382193 in ?? () from py/.dlis/lib/python3.8/site-packages/dlisio/core.cpython-38-x86_64-linux-gnu.so
#10 0x00000000005f1625 in PyCFunction_Call ()
[...]

PR #279, opened 2020-07-07
Handle files that are corrrupted by being zeroed-out before EOF

ejschoen:
This handles the issue I raised in #278:

- dl::findoffsets returns early if the length of the next logical record is 0.
- dl::stream_offsets gets a field called "stopped_early," which is set to true if the above happens.  (Arguably, this could be called "is_corrupted.")
- dlisio.Batch gets an optional property is_corrupted, set by keyword parameter in constructor.  Set to True when stream_offsets.stopped_early is true.  

dlisio.load returns whatever it was able to process.  The calling program can test for corruption:
```
   files = dlisio.load(filename)
   application_specific_processing(files)
  if (files.is_corrupted): logger.warning("Corrupted!")
```


ErlendHaa:
Hi, thanks for the fix!

I think the general idea here is good. Returning an error-code from `findoffsets` seams to be the reasonable choice as it gives the caller the freedom to decide how to precede. 

However, with such implementation I would like `findoffsets` to do some more sanity checking of the correctly read headers in order to supply a more granular error-code. E.g. we can determine not only that the logical file is truncated, but also whether the _last_ logical records is truncated or not.  Then there is a question about robustness; should the padding start right after a full VR (not _in_ one like here) then the `read_header_from_disk` from the `rp66`-protocol (or tapeimage) in `lfp` would fail. So checking `len` from `lrsh` is just a partial fix of the general problem that is padded-files. 

Until we have a good model for dealing with padding I would prefer to simply throw an exception whenever `len < 4`, to fix the infinite loop.

PR #280, opened 2020-07-13
Move implementation of the metadata pool to core

ErlendHaa:


python/dlisio/plumbing/linkage.py
```diff
@@ -5,28 +5,28 @@
...
     if not isinstance(obj, core.objref): raise TypeError
-    return obj.fingerprint, obj.type
+    return obj.name, obj.type
 
 def lookup(lf, reftype, value):
     """Create a fingerprint from reftype(value) and look up corresponding
     object in the logical file."""
     try:
-        fp, objtype = reftype(value)
+        name, objtype = reftype(value)
```
achaikou, line 19: maybe call it obname to hint at the type?

lib/src/io.cpp
```diff
@@ -494,7 +495,19 @@ noexcept (false) {
         dl::obname tmp{ dl::origin{ origin },
                         dl::ushort{ copy },
                         dl::ident{ std::string{ id, id + idlen } } };
-        xs.emplace_back(tmp.fingerprint("FRAME"), iflr.tell);
+        std::string fp = tmp.fingerprint("FRAME");
+
+        /* Although index.size() often is fairly large, the number of
+         * unique frames are typically just an handful. Hence the repeated
```
achaikou, line 501: number of unique frames (is) typically (small)
or
there are just (a) handful of unique frames?

number is a handful sounds strange to me

lib/src/io.cpp
```diff
@@ -494,7 +495,19 @@ noexcept (false) {
...
+         * calls to find() are not so bad as it might seem at first glace.
```
achaikou, line 502: I also want glace!!!! But probably we have to be satisfied with gla(n)ce :crying_cat_face:

python/tests/test_object_structures.py
```diff
@@ -5,6 +5,7 @@
 from datetime import datetime
 import numpy as np
 
+import pytest
```
achaikou, line 8: Remove pytest, make codacy happy :D

lib/src/parse.cpp
```diff
@@ -989,19 +991,48 @@ const char* parse_set_component( const char* cur,
...
+    auto objs = parse_objects(tmpl, this->type, cur, end);
+
+    this->tmpl = tmpl;
+    this->objs = objs;
+
+    this->parsed = true;
+}
+
+bool object_set::isparsed() const noexcept (true) {
+    return this->parsed;
```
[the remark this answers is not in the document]
ErlendHaa: Yes. And also, what are you really proposing here? `std::vector< dl::basic_object > obj = nullptr` is not a valid expression

PR #291, opened 2020-08-26
Move the pool of metdata-objects to core

ErlendHaa:


python/dlisio/__init__.py
```diff
@@ -255,18 +222,8 @@ def __getitem__(self, type):
...
-            for rec, t
-            in zip(self.attic, self.record_types)
-            if t == type
-        ]
-
-        self.load(recs, reload=False)
-
-        return self.indexedobjects[type]
+        objs = self.metadata.match(type, core.exactmatcher())
+        return { x.fingerprint : x for x in self.promote(objs) }
```
achaikou, line 226: You promised to change it to something akin to
```
        objs = self.match('.*', type, core.exactmatcher())?
        return { x.fingerprint : x for x in objs }
```
If you want to use two-arguments match, that means we need to do something with design and implementation of "match" function to allow empty pattern.

python/dlisio/__init__.py
```diff
@@ -357,27 +314,18 @@ def unknowns(self):
...
-
-        self.load(recs, reload=False)
-
-        unknowns = defaultdict(list)
-
-        for key, value in self.indexedobjects.items():
-            if key in self.types: continue
-            unknowns[key] = value
+        unknowns = defaultdict(dict)
+        for t in set(self.metadata.types):
```
achaikou, line 322: Would be cool to have tests of kind
types = [Channel, My_set, Frame, My_set] 
to see that all objects from My_set are present

python/dlisio/ext/core.cpp
```diff
@@ -962,6 +1018,27 @@ PYBIND11_MODULE(core, m) {
...
     });
 
+    m.def( "metadata_pool", []( dl::stream& s,
+                                const std::vector< long long >& tells ) {
+        std::vector< dl::object_set > tmp;
+        for (const auto& tell : tells) {
+            auto rec = dl::extract(s, tell);
+            if (rec.isencrypted()) continue;
+            if (rec.data.size() == 0) continue;
+            tmp.push_back( dl::object_set( rec ) );
```
achaikou, line 1028: If you want to leave parse_objects, can this method use it?

python/dlisio/plumbing/channel.py
```diff
@@ -86,16 +86,32 @@ class Channel(BasicObject):
...
+            msg = 'Unable to lookup frame, {} has no logical file'
+            logging.info(msg.format(self))
+            return None
+
+        # Find the frame(s) that are claiming ownership over this channel
+        frames = []
+        for frame in self.logicalfile.frames:
+            for channel in frame.channels:
+                if channel.attic.name != self.attic.name: continue
+                frames.append(frame)
```
achaikou, line 102: (Might be good to test on frames containing channels with the same name but different origins. If agreed, let me know; I'll create a file)

lib/src/parse.cpp
```diff
@@ -1003,36 +1003,33 @@ const noexcept (false) {
...
-        parse_set_component(b.data(),
-                            b.data() + b.size(),
+object_set::object_set(dl::record rec) noexcept (false)  {
+        parse_set_component(rec.data.data(),
+                            rec.data.data() + rec.data.size(),
                             &this->type,
                             &this->name,
                             &this->role);
-        this->buffer = std::move(b);
+        this->record = std::move(rec);
```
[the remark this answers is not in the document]
ErlendHaa: > This code will transfer record ownership and memory management to object set?

yes

> a reference/pointer to record 

Well, who handles the lifetime of that object then? Also, deleting the record would result in a dangling pointer  ...

achaikou (commented):
As usual force-pushes are punished with more comments :D

lib/extension/dlisio/ext/types.hpp
```diff
@@ -390,6 +390,16 @@ using value_vector = mpark::variant<
     std::vector< units  >
 >;
 
+struct record {
```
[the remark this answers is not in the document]
ErlendHaa: One of the next thing on the list: Nuke dlis extinction. That will trigger a re-think of the current file structures

achaikou (commented):
Case dismissed, you've been proved innocent, merge at will!

PR #294, opened 2020-09-10
Make our handling of encoded strings more robust

ErlendHaa:


lib/src/io.cpp
```diff
@@ -436,10 +437,10 @@ stream_offsets findoffsets( dl::stream& file) noexcept (false) {
     return ofs;
 }
 
-std::vector< std::pair< std::string, long long > >
+std::map< dl::ident, std::vector< long long > >
```
achaikou, line 440: Commit 48ea8d2c36b40d8bcb3dfbf83634eb1081a1f11b doesn't compile without chnages from b0f208e65714bdb282b83c62ee4b8c24bebdbbb3

python/dlisio/plumbing/basicobject.py
```diff
@@ -240,10 +240,10 @@ def fingerprint(self):
...
         fingerprint : str
         """
-        return core.fingerprint(self.type,
-                                self.name,
-                                self.origin,
-                                self.copynumber)
+        return core.fingerprint(self.attic.type,
+                                self.attic.name.id,
+                                self.attic.name.origin,
+                                self.attic.name.copynumber)
```
achaikou, line 246: Are these changes still relevant?
They are the same, but the attic version seems more internal than without attic

achaikou (commented):
:+1: 
[cosplaying Odysseus]:
None in here is non-nice!

python/tests/test_encodings.py
```diff
@@ -163,4 +163,42 @@ def test_broken_utf8_set(tmpdir, merge_files_oneLR):
...
+
+        # However it can be found by matching the bytes
+        with pytest.warns(UnicodeWarning):
+            obj = f.object('VERY_MUCH_TESTY_SET', b'\xeb\xe1\xe4\xf2')
+
+        assert obj.name == b'\xeb\xe1\xe4\xf2'
+
+        # When the encoding of the string parameter matches the encoding dlisio
+        # uses, we should expect to find the object
+        dlisio.set_encodings(['koi8_r'])
```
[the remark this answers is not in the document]
ErlendHaa: I'll update the docs, but will leave the test

python/dlisio/plumbing/basicobject.py
ErlendHaa: It's more robust to read them from c++. Any encoding changes will take place right away, while using the cached version means the object must be reloaded. Yes, I know we will not (at least for now) be able to support that feature for all strings (fdata_index). But we should improve where we can. Ideally I want to kill the cached name and type in these objects too. But that means updating a lot of tests I think, so that goes in the backlog

python/dlisio/dlisutils.py
```diff
@@ -10,7 +10,11 @@ def curves(dlis, frame, dtype, pre_fmt, fmt, post_fmt):
     Reads curves for provided frame and position defined by frame format:
     pre_fmt (to skip), fmt (to read), post_fmt (to skip)
     """
-    indices = dlis.fdata_index[frame.fingerprint]
+    try:
+        indices = dlis.fdata_index[frame.fingerprint]
+    except KeyError:
+        indices = []
```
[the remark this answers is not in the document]
ErlendHaa: It always has been like that. Before we uses a defaultdict(list). If we change it, we change the behaviour of frame.curves() for all frames without data

lib/src/io.cpp
```diff
@@ -466,7 +467,18 @@ noexcept (false) {
...
+        /* Although index.size() often is fairly large, the number of
+         * unique frames are typically just an handful. Hence the repeated
+         * calls to find() are not so bad as it might seem at first glace.
+         */
+        const auto itr = xs.find(fp);
+        if (itr == xs.end()) {
+            xs.insert( { std::move(fp), { tell } });
+        } else {
+            itr->second.push_back( tell );
+        }
```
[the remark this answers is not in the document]
ErlendHaa: You are right, and that should be fine:

> Reference to the mapped value of the new element if no element with key key existed. Otherwise a reference to the mapped value of the existing element whose key is equivalent to key.

https://en.cppreference.com/w/cpp/container/map/operator_at

python/dlisio/plumbing/matcher.py
```diff
@@ -42,6 +42,7 @@ class exact_matcher(core.matcher):
     def __init__(self):
         core.matcher.__init__(self)
 
-    def match(self, pattern, candidate):
+    @staticmethod
```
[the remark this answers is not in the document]
ErlendHaa: Because it's not an instance-specific method (read we don't use self). Although I can't find any notes about it in pybind11's docs, the macro we use in the trampoline apparently do handle it correctly.

lib/src/io.cpp
ErlendHaa: What, no? I just ran `git rebase -i --exec="cd build && make -j6 && ctest --output-on-failure" 1483e32` without any issues

achaikou: Why 1483e32 when the problem is in 48ea8d2?

ErlendHaa: That means, pick every commit _from_ 1483e32 to HEAD and run `cd build && make -j6 && ctest --output-on-failure`.

python/dlisio/plumbing/matcher.py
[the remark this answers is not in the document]
achaikou: Not sure that I like it (something seems suspicious), but ok...

achaikou (commented):
Rebase, fix fixups and I am fine with merging :+1:

PR #296, opened 2020-09-24
Error handling

achaikou:
- Performance seems to be mostly fine, checked one correct 3.4GB file once, process time increased from 1m 38s to 1m 40s, so within error margin.
- Real files with representation code issue seem to also behave as expected (I get the error about attributes in the logs, all other parameters in the set are processed).


ErlendHaa (commented):
Good work. I think this is the pr that will close the most issues at once this year!

python/dlisio/plumbing/errorhandler.py
```diff
@@ -4,10 +4,124 @@
...
 class ErrorHandler(core.error_handler):
-    """
-    TODO ErrorHandler
-    """
+    """ Defines rules about error handling
+
+    Many .dlis files happen to be not compliant with specification or simply
+    broken. This class gives user some control over handling of such files.
+
+    For example:
```
ErlendHaa, line 12: Rather than listing some examples here, I would spell out _all_ the different categories (critical, major, minor, info) and what the generally mean, default behaviour and maybe an example. Without a definition lines like 14 "... dlisio doesn't consider critical..."  makes no sense. I read that and think: but what makes dlisio consider something to be critical??  

>When dlisio encounters a spec violation it categories the issue based on the severity of the violation. Some issues are easily ignorable while other might force dlisio to give up on its current task. `ErrorHandler` supplies an interface for changing how dlisio react to different violation in the file. The different categories are `info`, `minor`, `major` and `critical`:

E.g critical: 

> Any issue that forces dlisio stop its current objective prematurely is categorised as `critical`. By default a `critical` error raises a `RuntimeError`. An example would be file indexing, which happens at load. Suppose the indexing fails midways through the file. There is no way for dlisio to reliably keep indexing the file. However, it is likely that the file is readable up until the point of failure. Changing the behaviour of `critical` from raising an Exception to logging would in this case mean that a partially indexed file is returned by load.

python/dlisio/plumbing/errorhandler.py
```diff
@@ -4,10 +4,124 @@
...
+
+    * Users can throw on specification violations which dlisio doesn't consider
+      critical.
+    * Users can escape errors if they want to read as much data as possible.
+
+    Error handler does not cover all the warnings and exceptions.
+    It is mostly intended to be used for data processing, not for data
```
ErlendHaa, line 19: Im no huge fan of the word processing. Everywhere else dlisio is categorised as a parser or dlis-reader. Now suddenly its a processor? Please don't confuse people with new words if not necessary. (read or parse are not inherently better than processing, but we don't need yet another pseudo word in the mix)

python/dlisio/plumbing/errorhandler.py
```diff
@@ -4,10 +4,124 @@
...
+    only changes default behavior.
+
+    By supplying :py:attr:`Actions.RAISE` for lesser violations, user can
+    strengthen the behavior and throw even on minor issues.
+
+    By supplying :py:attr:`Actions.LOG_ERROR` for critical issues, user relaxes
+    the behavior, thus allowing errors escape.
+
+    By default only CRITICAL violations would be thrown as errors, everything
+    else will be logged.
```
ErlendHaa, line 44: If anything, these are examples. However if you do something akin the critical example above, then this is sorta backed into that, so not need to spell it out explicitly

python/dlisio/plumbing/errorhandler.py
```diff
@@ -4,10 +4,124 @@
...
+    user must be careful when using this mode during close inspection.
+    If user decides to accept errors, they must be aware that some returned data
+    will be spoiled. Most likely it will be data which is stored in the file
+    near the failure.
+
+    Warnings
+    --------
+    If user decides to accept errors, dlisio will return all the data that it
+    was able to process. But a lot of data after the failure may be skipped as
+    usually errors are not recoverable.
```
ErlendHaa, line 74: This too is obvious and not needed if you properly define a critical error in the first place

python/dlisio/plumbing/errorhandler.py
```diff
@@ -4,10 +4,124 @@
...
+    Warnings
+    --------
+    Be careful not to ignore too much information when investigating files.
+    dlisio might classify some issue as a minor violation, but it might be
+    critical for understanding the true level of data brokeness.
+
+    Notes
+    -----
+    Users do not have direct access to issues found during processing. They are
+    reported on user's behalf in logger or directly thrown as an exception.
```
ErlendHaa, line 85: What does this mean exactly, and is this information that the user need to know?

python/dlisio/plumbing/errorhandler.py
```diff
@@ -4,10 +4,124 @@
...
+    --------
+    If user decides to accept errors, dlisio will return all the data that it
+    was able to process. But a lot of data after the failure may be skipped as
+    usually errors are not recoverable.
+
+    Warnings
+    --------
+    Be careful not to ignore too much information when investigating files.
+    dlisio might classify some issue as a minor violation, but it might be
+    critical for understanding the true level of data brokeness.
```
ErlendHaa, line 80: Definition of `minor`?

python/dlisio/plumbing/errorhandler.py
```diff
@@ -4,10 +4,124 @@
...
+    ...        minor    = dlisio.ErrorHandler.Actions.LOG_WARNING,
+    ...        major    = dlisio.ErrorHandler.Actions.RAISE,
+    ...        critical = dlisio.ErrorHandler.Actions.RAISE)
+
+    Process a file to have a general overview:
+
+    >>> try:
+    ...    with dlisio.load(path) as fs:
+    ...        res = process(fs)
+    ...        store_results(path, res, broken=False)
```
ErlendHaa, line 102: this fictional functions through me of, and I'm a developer of this thing. They are beside the point, so just do `pass`?  I think a better illustration of the behavior is to drop the the try and let it fail, then apply custom a handler and see how everything works:

```
>>> files = dlisio.load(path)
RuntimeError: "...."

>>> handler = ErrorHandler(critical=LOG_ERROR)
>>> files = dlisio.load(path)
[ERROR] "...."
>>> for f in files:
...  pass
```

Showing that people can change behaviour midways is to unnecessarily complicate things I think. The one expert user can figure that one out for himself. But maybe add the a small docstring on the handler attribute in logical_file, that should be a hint to anyone digging around that this can be changed at any point.

python/dlisio/ext/core.cpp
```diff
@@ -1083,11 +1083,20 @@ PYBIND11_MODULE(core, m) {
...
+                                dl::error_handler& errorhandler ) {
         std::vector< dl::object_set > objects;
         for (const auto& rec : recs) {
             if (rec.isencrypted()) continue;
-            objects.push_back( dl::object_set( rec ) );
+
+            try {
+                objects.push_back( dl::object_set( rec ) );
+            } catch (const std::exception& e) {
+                const auto context = "load (object set creation)";
```
ErlendHaa, line 1095: I find all these `context = load`a bit weird. Yes they are generally used from load, but doesn't have to be. Maybe do, `context = "{function} {what the function essentially does}"`? Like here: `context = "core.parse_objects: Construct object sets"`. 

This also applies to findoffsets, findfdata, readfdata, extract and object_set.parse.

Sidenote, this functions doesn't really parse objects does it? It just initialises object set

python/dlisio/ext/core.cpp
```diff
@@ -698,13 +700,33 @@ noexcept (false) {
...
+        record_dst = dst;
+
         /* get record */
-        auto record = dl::extract(file, i);
+        dl::record record;
+        try {
+            record = dl::extract(file, i);
+        } catch (std::exception& e) {
+            handle(e.what());
+            continue;
```
ErlendHaa, line 724: At the start of read_fdata we allocate rows equal to the number of fdata. What happens to the array when we don't fill it up completely because we skip some frames? Should it be resized?

ErlendHaa (commented):
Great work!

achaikou (commented):
Thanks for the thorough review! :+1: 
Majority is fixed.
I am not arguing with the docs deity :smile:

python/dlisio/plumbing/errorhandler.py
```diff
@@ -0,0 +1,83 @@
...
+        if specification:
+            spec_format = "\nSpecification reference: {}."
+            spec_formatted = spec_format.format(specification)
+
+        action_formatted = ""
+        if action:
+            action_format = "\nTaken action: {}."
+            action_formatted = action_format.format(action)
+
+        msg = "{}{}{}{}"
```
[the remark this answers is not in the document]
ErlendHaa: 1) Would put severity after problem. It's not exactly the essence of the msg, so putting it first seams a little wierd.
2) Would also put problem first:
```
problem
where
severity
spec
action
```
Then the most important information comes first
3) would drop the periods at the end of each line. Looks better imo,  and it seam to be a convention. I see no other errors (in dlisio or else where) that uses periodes

lib/src/parse.cpp
```diff
@@ -854,10 +878,19 @@ noexcept (false)
         case rpc::status: reset< dl::status >(value).resize(count); return;
         case rpc::units:  reset< dl::units  >(value).resize(count); return;
         default: {
-            const auto msg = "unable to patch attribute with no value: "
-                             "unknown representation code {}";
+            // repcode is incorrect, but value is missing
+            // hence we can log an error but continue processing
+            const auto msg =
+                "unable to interpret value, invalid representation code {}";
```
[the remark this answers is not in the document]
ErlendHaa: My bad, I was wrong the first time

lib/src/parse.cpp
```diff
@@ -892,12 +892,21 @@ noexcept (false)
     }
 }
 
+bool is_log_clear( const std::vector< dlis_error >& log ) noexcept (false) {
+    for (const auto& entry : log) {
```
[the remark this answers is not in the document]
ErlendHaa: Nope, just an observation from my side.

lib/src/parse.cpp
```diff
@@ -896,6 +901,19 @@ bool is_log_clear( const std::vector< dlis_error >& log ) noexcept (false) {
     return true;
 }
 
+void report_set_errors(const dl::object_set& eflr,
+                       const dl::error_handler& errorhandler) {
+    if (eflr.log.size()) {
+        const auto context = "object set of type '" +
```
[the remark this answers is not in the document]
ErlendHaa: Yep, stupid me forgot

python/dlisio/plumbing/basicobject.py
```diff
@@ -196,6 +196,17 @@ def __getitem__(self, key):
             # No rule for parsing, keep rp66value as is, i.e. vector
             parse_as = vector
 
+        # report errors before checking for key presence - it might be a symptom
+        if len(self.attic.log) > 0:
+            # TODO: here and for attribute: we use fingerprint to report
```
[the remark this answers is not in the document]
ErlendHaa: Yes, I meant to change `return '{}({})'.format(self.type.capitalize(), self.name)`. But next pr

lib/src/io.cpp
```diff
@@ -405,23 +405,44 @@ record& extract(stream& file, long long tell, long long bytes, record& rec) noex
...
+            file.read(buffer, DLIS_LRSH_SIZE);
+        } catch (std::exception& e) {
+            handle(e.what());
+            break;
+        }
+
         if (file.eof()) {
-            check_for_truncation(file, offset);
+            if (is_truncated(file, lrs_offset)) {
+                const auto problem = "findoffsets: file truncated";
```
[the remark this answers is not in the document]
ErlendHaa: But probably not much. Have you measured this? Im not so sure it will have a noticeable impact. By reading the last byte of each segment what we essentially do is read _5 successive bytes_ instead of 4. But it would be good to measure anyway

lib/src/io.cpp
```diff
@@ -430,44 +451,72 @@ stream_offsets findoffsets( dl::stream& file) noexcept (false) {
...
+
+        if (not (attrs & DLIS_SEGATTR_SUCCSEG)) {
+            /*
+             * As we allow to return from findoffsets correctly even if file
+             * is broken, we need to make sure that offsets contain only
+             * fully-processed records. Hence check for truncation before
+             * storing a LR on last processed segment
+             */
+            if (is_truncated(file, lrs_offset)) {
+                const auto problem = "findoffsets: file truncated on last lrs";
```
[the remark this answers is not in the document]
ErlendHaa: I see you moved it, so my previous comments doest fit anymore. But it's still the record segment that is truncated. So maybe the message should still be something akin to `Record segment truncated`

lib/src/io.cpp
```diff
@@ -405,23 +405,44 @@ record& extract(stream& file, long long tell, long long bytes, record& rec) noex
     }
 }
 
-stream_offsets findoffsets( dl::stream& file) noexcept (false) {
+stream_offsets findoffsets( dl::stream& file, dl::error_handler& errorhandler)
```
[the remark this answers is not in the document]
ErlendHaa: Probably one of them, and probably that should be lr_offset! The hard-core debugger then has enough information to find the underlying tell herself
```
io.seek(lr_offset)
io.absolute_tell()
```

python/tests/test_error_handler.py
```diff
@@ -16,7 +16,41 @@
...
 # TODO: test mix (do not fail on truncation, but fail on bad attribute access)
 
-# TODO: test_truncated
+def test_unescapable_notdlis(assert_error):
+    path = 'data/chap2/nondlis.txt'
+    with pytest.raises(RuntimeError) as excinfo:
+        _ = dlisio.load(path, error_handler=errorhandler)
+    assert "could not find visible record envelope" in str(excinfo.value)
+
+def test_truncated_in_data(assert_error):
```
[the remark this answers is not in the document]
ErlendHaa: I mean if dlisio behave differently if the truncations hits in tape image protocol first?

python/dlisio/ext/core.cpp
ErlendHaa: You are completely right, my bad

lib/src/parse.cpp
```diff
@@ -944,31 +1032,54 @@ object_vector parse_objects( const object_template& tmpl,
                     const auto msg = "count ({}) isn't 0 and representation "
                         "code ({}) changed, but value is not explicitly set";
                     const auto code = static_cast< int >(attr.reprc);
-                    user_warning(fmt::format(msg, count, code));
+                    dlis_error err {
+                        dl::error_severity::MAJOR,
+                        fmt::format(msg, count, code),
+                        "-",
```
ErlendHaa, line 1038: remove `-`

python/dlisio/plumbing/errorhandler.py
```diff
@@ -0,0 +1,83 @@
...
+    @staticmethod
+    def swallow(msg):
+        pass
+
+
+    class Actions(Enum):
+        """
+        TODO Actions
+        """
+        LOG_DEBUG   = lambda msg: logging.debug(msg)
```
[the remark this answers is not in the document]
ErlendHaa: Yes I commented on the first commit where they where introduced, and forgot to re-visit my comment when the docs later appeared. My bad

lib/src/parse.cpp
```diff
@@ -999,10 +1009,32 @@ const char* object_set::parse_objects(const char* cur) noexcept (false) {
...
             }
 
+            object_clear = object_clear && is_log_clear(attr.log);
+
             current.set(attr);
         }
 
+        if (not object_clear) {
+            const auto msg =
+                "Some object attributes violate specification, which might "
```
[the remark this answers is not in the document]
ErlendHaa: 1. I agree that my suggestion was not the best. But I'm still unsure if the current message is clear enough. Maybe we should simply say `This object is a part of a object-set that violated the specification`. Then as this is just info, we don't explicitly need to say how trustworthy it is?
2. We also set a similar message for the objects-set. Currently we log the set-level when querying E.g. `f.object`, `f.find`. Then when someone tries to _use_ the object, we log the object-level msg. Consider the normal use case:
```
obj = f.object(...)
_ obj.attribute
```
that will issue two almost identical messages, but even to me it's not completely obvious that they refer to the _exact_ same issue.
```INFO     root:errorhandler.py:141 
Severity:     minor.
Where:        object set of type 'VERY_MUCH_TESTY_SET' named 'SETNAME'.
Problem:      Some objects in this object set violate specification, which might mean that correctness of other objects is also affected.
INFO     root:errorhandler.py:141 
Severity:     minor.
Where:        T.VERY_MUCH_TESTY_SET-I.OBJECT-O.1-C.1.
Problem:      Some object attributes violate specification, which might mean that correctness of other attributes is also affected. 
```
So my question is, do we need to set this information on the set-level? Does it bring any value or is it just extra noise and confusion? I'm leaning towards the last. And I don't see a big problem with the user being delivered a "broken" object without a warning, as long as the warning comes when they try to use the object. After all, that's how we deal with all other objects.

(This time I'm not arguing _where_ the set-level messages should be issued from, just it _this specific_ message is needed)

python/dlisio/plumbing/errorhandler.py
```diff
@@ -0,0 +1,216 @@
...
+             By default a critical error raises a RuntimeError.
+
+             An example would be file indexing, which happens at load. Suppose
+             the indexing fails midways through the file. There is no way for
+             dlisio to reliably keep indexing the file.
+             However, it is likely that the file is readable up until the point
+             of failure. Changing the behaviour of critical from raising an
+             Exception to logging would in this case mean that a partially
+             indexed file is returned by load.
+    major    Any issue that dlisio is able to recover from under heavy
```
ErlendHaa, line 35: > heavy developer's assumption

Generally we refer to dlisio as the entity in docs, not to the developers. It's implied that any behaviour dlisio has is decided by developers. I also generally try to avoid formulating sentences such that I have to use the word `user`

Also, heavy assumptions. That sounds frightening, and the definition of `heavy` in this case is up for grabs. rather than trying to define heavy and light I would try to explain the behaviour of dlisio when a major/minor issue is encountered.

major:

These are a result of direct spec violations in the file, but where dlisio makes an assumption on what the broken information [1] _should_ have been, and continues parsing the file on this assumption. If no other major or critical issues are reported, it's likely that the assumption where correct and that dlisio parsed the file correctly. However, no guarantees can be made.

[1] Note that "information" in this case refers to data in the file that tells dlisio how the file should be parsed. Not to the actual parsed data.

Minor

Like, Major these are also a result of direct spec-violations, and dlisio makes similar assumptions to keep parsing the file. Minor issues are generally less severe and easier to handle correctly. However, no guarantees can be made about the parsed data

python/dlisio/plumbing/errorhandler.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+    Warnings
+    --------
+    Escaping errors is a good solution when user needs to read as much data as
+    possible, for example, to have a general overview over the file. However
+    user must be careful when using this mode during close inspection.
+    If user decides to accept errors, they must be aware that some returned data
+    will be spoiled. Most likely it will be data which is stored in the file
+    near the failure.
```
ErlendHaa, line 97: Maybe this is too obvious to mention specifically?

python/dlisio/plumbing/errorhandler.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Warnings
+    --------
+    Be careful not to ignore too much information when investigating files.
+    dlisio might classify some issue as a minor violation, but it might be
+    critical for understanding the true level of data brokeness.
```
ErlendHaa, line 103: What is "the true level of data brokenness", and when do I need to understand it?

I think what you are essentially trying to say is that "if you want to debug a broken part of the file, you should look at _all_ errors and warnings (from that part) to get a full picture of the situation." Or am I mistaken?

lib/src/io.cpp
```diff
@@ -388,67 +413,157 @@ record& extract(stream& file, long long tell, long long bytes, record& rec) noex
...
+        } catch (std::exception& e) {
+            handle(e.what());
             break;
+        }
+
+        // read operation is enough to set eof correctly
+        if (file.eof()) {
+            if (read == 0) {
+                if (has_successor) {
+                    auto problem =
```
ErlendHaa, line 448: const auto. (There are more in findoffsets)

lib/src/io.cpp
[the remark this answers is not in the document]
achaikou: As per our investigation just combined seeks.

lib/src/io.cpp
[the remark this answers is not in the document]
achaikou: I think I just unified all the situations under "File truncated"? 

Both messages mean the same and it seems like "file" is a bit more user-friendly?

python/dlisio/plumbing/errorhandler.py
achaikou: 1. https://i.redd.it/nx8b897mg9g41.png
2. I wouldn't say that it's too obvious. If I try imaging myself as a dlisio user, there is no logical connection for me between concepts "I accepted errors" and "Something might be wrong in the data that I received"
In my mind it's "Error is reported at line 11, it means that I can trust lines 1-10", which doesn't necessary hold in case of, for example, last returned record of indirectly formatted data. If data itself is spoiled, we will never know.

lib/src/io.cpp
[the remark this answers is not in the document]
ErlendHaa: Message should be `file truncated in Logical Record`, otherwise all good

lib/src/io.cpp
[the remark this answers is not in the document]
ErlendHaa: A hybrid? `file truncated in logical record segment`. User-friendly, and specific

python/dlisio/errors/errorhandler.py
```diff
@@ -189,26 +189,26 @@ def format_severity(severity):
     # TODO: nested format
     @staticmethod
     def format_error(severity, context, problem, specification, action):
-        severity_format = "\nSeverity:     {}."
-        severity_formatted = severity_format.format(
-            ErrorHandler.format_severity(severity))
+        problem_format = "\nProblem:      {}"
+        problem_formatted = problem_format.format(problem)
```
ErlendHaa, line 193: pro tip on string-formatting alignment

```
format = '\n{:<{align}} {}'
problem_formatted = format('Problem:', problem, align=10) # Or whatever is the max 'key' length is
```

python/dlisio/plumbing/errorhandler.py
ErlendHaa: 1) hahaha
2) fine :D

lib/src/parse.cpp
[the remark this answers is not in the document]
ErlendHaa: Im still not to fuzzed about this error msg, but other than that I'm happy

> some object attributes violate specification, which might mean that correctness of other attributes is also affected

It's to wage for me, and that causes confusion and a bit of frustration when reading it. (why don't I understand this, it seams like something I such understand as there is no unfamiliar concepts)
> some object attributes 

what does that means, from _this_ object, from another object somewhere else, can a parameter attribute fuck up a channel attribute like this? 

> correctness of other attributes

Which attributes are we talking about now, the once on _this_ object, or is it more general?

In this case we have the option to be either too specific (e.g. users generally don't know what a set is) or too wage which cases equally, if not more confusion. 

I would say:
"One or more attributes on this object violated the spec, this can potentially corrupt the entire object" 
Or something to that effect

lib/src/parse.cpp
```diff
@@ -1139,15 +1139,31 @@ object_set::object_set(dl::record rec) noexcept (false)  {
...
+            "object set parse has been interrupted"
+        };
+        this->log.push_back(err);
+    }
+    // TODO: possible assert that cur == end of data
+
+    /* If set is parsed in default mode, exception will be thrown and set won't
+     * be considered parsed. If set is parsed after that in error-escape mode,
+     * parsing will be locked, but error will get stored on object set anyway.
+     */
```
ErlendHaa, line 1174: Is this still valid? I don't understand it

python/dlisio/errors/errorhandler.py
```diff
@@ -0,0 +1,206 @@
...
+    @staticmethod
+    def format_severity(severity):
+        if   severity == core.error_severity.info:     return "info"
+        elif severity == core.error_severity.minor:    return "minor"
+        elif severity == core.error_severity.major:    return "major"
+        elif severity == core.error_severity.critical: return "critical"
+        else:                                          return severity
+
+    # TODO: check if that helps for named fileheader issue
+    # TODO: nested format
```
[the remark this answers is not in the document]
achaikou: Yes. I still need to check both!

lib/src/parse.cpp
achaikou: Yes, it is valid.
What I mean that parsing with escaping exceptions locks your from getting exceptions ever again.

Lets say you do next:
1. parse without escaping errors -> exception
2. parse again without escaping errors -> exception
3. parse with escaping errors -> parsed
4. parse again without escaping errors - > parsed

PR #300, opened 2020-10-15
alternative approach for retrieving parameter units

eimerej:
Here is an alternative approach to retrieve parameters units by creating a fake Unit attribute.
Of curse again, it is just a draft.


ErlendHaa:
Hi, as we opted for the approach suggested in #299 (and merged as a part of #302), I'm going to close this issue now

PR #306, opened 2020-10-27
Add logicalfile.find and deprecate match, __getitem__

ErlendHaa:


achaikou (commented):
You were right :smile: Seems like I like this one too :+1:

python/dlisio/settings.py
```diff
@@ -1,4 +1,17 @@
...
+
+import re
+
+
+""" regex and exact matchers are frequently used by most methods on
+logicalfile. To avoid the overhead of initializing a new instance of these for
+every method-call they are cached as globals here.
+
+Although possible, these globals are not intended to be changed by the end user directly.
+"""
```
achaikou, line 12: Not sure if this is why Codacy complains, but maybe this doesn't have to be docs (as it refers to both settings), just a normal comment.
Or if we make it docs, it might have to be put under the line it refers to

achaikou (commented):
Looks good to me :+1: 
Merge while I am still nice :D

PR #310, opened 2020-12-02
Enrich describe

ErlendHaa:


python/dlisio/plumbing/parameter.py
```diff
@@ -137,10 +137,17 @@ def values(self):
...
+        try:
+            units = self.attic['VALUES'].units
+        except:
+            pass
+
         d = OrderedDict()
         d['Sample dimensions'] = self.dimension
         d['Axis labels']       = self.axis
         d['Zones']             = self.zones
+        d['Units']             = units
```
achaikou, line 150: I admit I saw it somewhat differently when I suggested it. In my mind outcome must be something like:

```
  Calibration-measurement
   -----------------------
   name   : MSCT/KPPO_PM/CALI_MEASUREMENT
   origin : 2
   copy   : 0

   Metadata
   --
   Phase in job sequence     : BEFORE
   Start time of acquisition : 2011-08-20 18:27:54
   Duration time             : 5 s

   Samples
   --
   Number of values                    : 1
   Samples used to compute std/max-dev : 12

   Value(s)      : 2.5799591541290283 in
   Reference     : 2.140000104904175 in
   Std deviation : 0.004975703079253435 in
   Max deviation : 0.010652303695678711 in
```
or

```
   ---------
   Equipment
   ---------
   name   : MSCT/MCFU_1/EQUIPMENT
   origin : 2
   copy   : 0

   Trademark name     : MCFU_1-AA
   Operational status : 1
   Serial number      : 119.

   Length         : 125.0 in
   Diameter (min) : 5.25 in
   Volume         : 1.5700000524520874 ft3
   Weight         : 144.0 lbm

   Max pressure          : 20000.0 psi
   Max temperature       : 350.0 degF
```

Not sure if that's the same issue you are wondering about if it's worth it, because I probably think it is.

It's not very fast to have it, but doesn't seem like too much code to me. Roughly updated 
- lines in objects to something like `d['Type of measurement']       = "TYPE"`
- do `describe_attributes(buf, d, self.attic, width, indent, exclude) `instead of dict
- updated `describe_array` and `describe_text` signatures to include units and actually print them
- changed `describe_dict` to `describe_tup_dict` (so value is (value, units))
- `describe_sampled_attrs` and `describe_attributes` retrieve units as they know attic and label

Obv, I am missing a lot of useful things, but still.

Besides, even if units were take from a proper place, we would still have to deal with it. So maybe now is better than later.

(The biggest change is all the labels. Maybe we should again think about 1-1 mapping between our attribute names and attic labels. Maybe we can do something about it so we always use self["LABEL"] inside of the code and leave self.label barely for users benefit)

python/dlisio/plumbing/frame.py
```diff
@@ -482,13 +482,16 @@ def fmtstrchannel(self, channel):
...
                 d = OrderedDict()
-                d['Description']      =  self.description
-                d['Indexed by']       =  self.index_type
-                d['Interval']         =  str([self.index_min, self.index_max])
-                d['Direction']        =  self.direction
-                d['Constant spacing'] =  self.spacing
-                d['Index channel']    =  self.channels[0]
+                d['Description']      = self.description
+                d['Indexed by']       = self.index_type
+                d['Units']            = index.units
```
achaikou, line 490: 1. Not convinced we should add them here at all.
2. Index units?
3. Why self.channels when we have .index method for where first channel isn't actually index one?
4. What will happen if frame is indexed by framenos?
5. Units from index channel are different from units written in the file. Example from our file:

```
   -----
   Frame
   -----
   name   : 800T
   origin : 2
   copy   : 0

   Channel indexing
   --
   Indexed by       : TIME
   Index min        : 33354518 0.5 ms
   Index max        : 35194520 0.5 ms
   Direction        : INCREASING
   Constant spacing : 800 0.5 ms
   Index channel    : Channel(TIME)

   Units from index : ms
```

So units actually are 0.5 ms in the file, whatever that means (half millisecond??) 
So if we print them, we probably need to print it somewhat separately as `35194520 [0.5 ms]`

python/dlisio/plumbing/utils.py
```diff
@@ -343,8 +343,12 @@ def describe_text(buf, text, width, indent, subindent=None):
     if subindent is None: subindent = indent
 
     if text is None: text = str(None)
+    else: text = str(text)
+
+    if len(text) == 0: text = '""'
```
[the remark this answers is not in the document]
ErlendHaa: 1) yes, when no content the list is empty. However, the __repr__ of an empty list is `[]`, while an empty string is just an empty string. Given an attribute `attribute` where the value is an empty dict, the output in describe (for that line) becomes:` attribute : []`, but if the value was string, and an empty one, this would output as a blank line. This commit makes an empty string behave like an empty list and the line now becomes `attribute : ""`
2) it only applies if you explicitly tell describe to not exclude empty attributes

python/dlisio/plumbing/parameter.py
ErlendHaa: I'm gonna bother with this now, as I do not think it's worth the effort.

python/dlisio/plumbing/frame.py
ErlendHaa: 1) Gonna ignore as you don't present any arguments for your claim
2) sure, but seems redundant as it's in a section about indexing
3) Can use that inside the `if`, but the `if`-statement itself does a manual check to avoid the potential logging from `.index`. 
4) no-index is handled exactly like before (`if-else`), no changes to that, and non needed.
5) Good catch, and that would technically be allowed, although stupid. Then I suggest

```
...
Index units : ms
Index min   : 33354518 [0.5 ms]
...
```

Then we do not miss-represent anything. And if the file doesn't make sense, it's not out problem

ErlendHaa: 3. I take it back, frame.index return the name of the index channel, not the index channel itself

ErlendHaa:
The commit from this pr is a part of #311

PR #318, opened 2021-01-11
Various

achaikou:


python/tests/test_loading.py
```diff
@@ -149,7 +149,14 @@ def test_context_manager_with():
 def test_load_nonexisting_file():
     with pytest.raises(OSError) as exc:
         _ = dlisio.load("this_file_does_not_exist.dlis")
-    assert "this_file_does_not_exist.dlis : No such file" in str(exc.value)
+    msg = "Path 'this_file_does_not_exist.dlis' is not an existing regular file"
```
ErlendHaa, line 152: Drop `Path`?

lib/src/dlis/io.cpp
```diff
@@ -336,8 +336,17 @@ noexcept (false) {
 
     const auto handle = [&]( const std::string& problem ) {
         const auto context = "dlis::findoffsets (indexing logical file)";
```
ErlendHaa, line 338: maybe avoid some string-creations?
```
const auto debug = "Absolute offset is {} (dec), "
                   "Logical Record tell is {}, "
                   "Logical Record Segment tell is {}"
                   
debug = fmt::format(debug, file.ptell(), lr_offset, lrs_offset);
```

What is dec ?

lib/src/dlis/io.cpp
```diff
@@ -514,6 +514,29 @@ dl::error_handler& errorhandler) noexcept (false) {
     return xs;
 }
 
+std::vector< dl::record >
```
ErlendHaa, line 533: I'm not convinced about this move. Here it is a part of the C++ API, and I don't think we want that. The purpose of this function is not _convenience_, it's _performance_. It avoids going back and forth between python and C++. I don't think we should expose this function for convenience, as people can write their own loops for their specific purposes.

lib/include/dlisio/exception.hpp
```diff
@@ -1,6 +1,8 @@
 #ifndef DLISIO_EXCEPTION_HPP
 #define DLISIO_EXCEPTION_HPP
 
+#include <stdexcept>
```
ErlendHaa, line 4: Also missing `#include <string>`

Issue #102, opened 2019-05-03
More debugging tools/information for dealing with issues in bytes processing


achaikou:
Let say DLISIO throws an error on opening some file.
If user wants to dig deeper into binary and figure out where is the problem, user might appreciate more information than is currently provided.
- It would be nice if user knew exact byte position where failure happened (or at least values of neighbour bytes to be able to identify failure position in file on their own).
- It would be beneficial to have easy access to already processed attributes in the same set, because actual failure might happen many bytes before reported place (imagine if bytes were incorrectly interpreted as 100-bytes long indent. Failure would be thrown at least 100 bytes after the actual error occurred.)

Hence it might be worth considering to extend support for binary debug.

achaikou:
More useful points for issues investigation: 

1. Include failure offset in error messages.
Make all the main methods (_findoffsets_, _stream.at_, _read_fdata_, parse functions, etc) return offset of the failure point. When figuring out what exactly is wrong with the data this is the main point for manual investigation (one needs to jump to the correct point in the file).
Major problem is however that this data is lost during high levels of processing.

2. Fail early if virtual record header doesn't contain bytes "FF01".
Currently _dlis_vrl_ completely ignores 'FF' bytes and majorly ignores version (01). If we process these bytes, we might fail early when they are not as expected (so this is not start of new VR). Right now we do nothing with them probably with consideration of other possible versions of dlis. However in all seen files bytes always were "FF01", so it might be beneficial to drop that all together.

3. Introduce the check "current logical record length (**len**) should be bigger than data remaining in VR (**remaining**)" in _findoffsets_. 
Without this check we often end up with negative value of **remaining** variable, which already is a clear indication that something went wrong.

PR #322, opened 2021-01-19
Q&A

achaikou:
3 sections that I felt are missing from documentation and which I myself would like to see as a user.

- Section for people who landed up on this project page but are totally clueless about what's going on and need to figure out if they should investigate dlisio further.
- Section for people who received one of our errors and are actually at loss with regards to what they should do about it. (aka the most common situation that I saw which lead to the provided error)
- Section for some future dlisio developers. Most likely no-one ever will find it useful, but as I already acquired some dlis debugging mindset, it probably makes sense to share it.

Feel free to completely annihilate small/big/~all~ parts of this PR. Mark everything that seems unclear on first read, it will likely be the same for users.

### DIFF
--- a/python/docs/beginner.rst
+++ b/python/docs/beginner.rst
@@ -1,0 +1,126 @@
+Overview questions
+==================
+
+Purpose of this section is to provide people who just stumbled upon dlisio with
+short overview.
+
+.. contents:: :local:
+
+General
+-------
+
+Is dlisio something I might be interested in?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+It might be useful for you if:
+
+1. you have some **.dlis** files you want to know contents of
+2. it is acceptable for you to use programming language to do so
+3. this programming language is `Python`_
+
+What do I need to know to work with dlisio?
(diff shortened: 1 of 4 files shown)


ErlendHaa:
Closes #102

achaikou:
To be updated some undefined time in the future:
- Move "Can I be sure that dlisio presents me the data correctly?" section to Readme
- Remove everything else from Q&A, thus dropping explanations for clueless repo viewers
- Add Errors explanation section under dlis, while removing everything that doesn't correspond to literal error string (aka performance messages).
- (Check that messages still make sense)
- Put investigation guide in the docs, but never link it publicly.

lucasblanes:
Hello! I don't know if this is the best place to ask, but can I create DLIS files using DLISio?

achaikou:
Hi @lucasblanes !

No, this is a place for proposed changes to the code. But I see why the title caused confusion :)
Better place to ask questions is [Issues](https://github.com/equinor/dlisio/issues) as it can be searched by others wondering the same.

To the question itself:
no, dlisio can't write dlis files. At the moment there are no plans to support it in any foreseeable future.

lucasblanes:
Hello @achaikou ! Sorry for the confusion and thank you for your answer!

PR #325, opened 2021-02-02
NOFORM and UDI support

### LINKED ISSUE #215: Consider supporting NOFORM and UDI
Refer to _5.9_ and _5.10_ parts of specification.
It's unclear how many files contain data like that and if users will be able to get any value from such data.

achaikou:
Closes #215


python/dlisio/plumbing/noformat.py
```diff
@@ -53,3 +54,74 @@ def describe_attr(self, buf, width, indent, exclude):
...
+        >>> out = open('noformat.txt', 'wb')
+        >>> out.write(noformat.data())
+        >>> out.close()
+
+        If you are dealing with images, some providers might have internal
+        hidden information.
+
+        >>> images = f.find('.*IMAGE.*', matcher=dlisio.plumbing.regex_matcher())
+        >>> for image in images:
+        >>>     print(image.describe())
```
ErlendHaa, line 93: What this example all about?

python/dlisio/plumbing/noformat.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def describe_attr(self, buf, width, indent, exclude):
+        d = OrderedDict()
+        d['Consumer name'] = 'CONSUMER-NAME'
+        d['Description']   = 'DESCRIPTION'
+
+        describe_attributes(buf, d, self, width, indent, exclude)
+
+    def data(self):
+        """All noform data belonging to object
```
ErlendHaa, line 59: Raw data

A raw byte buffer of all the records corresponding to the noform-object. This content of this data is not defined by rp66v1. However, the name and description of the no-format object might give an indication of what the content is

python/dlisio/plumbing/noformat.py
```diff
@@ -0,0 +1,127 @@
...
+        Consumer name : FILE
+        Description   : /home/files.txt
+
+        Save the data
+
+        >>> out = open('noformat.txt', 'wb')
+        >>> out.write(noformat.data())
+        >>> out.close()
+
+        If you are dealing with images, some providers might have internal
```
ErlendHaa, line 88: It is not uncommon that the no format object itself is not enough to understand its content. Some vendors store additional information in other custom object types. In that case some manual digging into the file might reveal the information necessary to correctly understand the no formats content:

ErlendHaa (approved):

PR #330, opened 2021-02-23
Refactor dlis.load

achaikou:
Intention of this PR was to make dlis.load flow as clear as possible.
We have a couple of totally valid .dlis files that dlisio failed to read correctly because they were "unlikely special case".
They should no longer be the problem.


python/dlisio/load.py
```diff
@@ -88,88 +88,209 @@ def load(path, error_handler = None):
...
+    # Processing logic:
+    # - Open cfile/tif protocol at the start of every logical file.
+    #   Note that TIF protocol may report EOF before physical file is EOF, thus
+    #   it's important to open it anew every time.
+    # - Read SUL if it is found.
+    # - Parse Logical file.
+    # - As TIF EOF marks can happen before SUL, before LF or before physical
+    #   EOF, test for EOF before attempting to read SUL or LF blocks.
+    # - Close streams that do not contain valid Logical Files.
+    # - Attempt to open logical files until physical file reports data end.
```
ErlendHaa, line 109: iho there are only a couple of things that are unclear about the while loop:

1) what's the deal with the `close + continue`
2) Where is what protocol opened
3) That parser.eof() also is true for truncated files

I would remove the current comments and add an inline comment about 1), properly explaining that this is a tif thing. As pr other comments I would address 2) by explicitly open rp66 in the loop (not hide it in parse_LF, and maybe rename open_stream. And point 3) by splitting EOF and truncated into separate attributes

python/dlisio/load.py
```diff
@@ -88,88 +88,209 @@ def load(path, error_handler = None):
...
+        explicits, implicits = self.find_offsets()
+        recs = core.extract(self.stream, explicits, self.error_handler)
+        sets = core.parse_objects(recs, self.error_handler)
+        pool = core.pool(sets)
+        fdata = core.findfdata(self.stream, implicits, self.error_handler)
+
+        lf = logicalfile(self.stream, pool, fdata, self.sul, self.error_handler)
+        self.lfs.append(lf)
+
+    def eof(self):
```
ErlendHaa, line 224: physical_eof? or peof?

lib/include/dlisio/stream.hpp
```diff
@@ -64,6 +64,16 @@ class stream {
...
+
+    /* Leaf protcol
+     *
+     * Gives user access to the most underlying layer of the stream.
+     *
+     * Note that this method must be used with extreme caution as
+     * both current and returned stream would work on the same underlying
+     * object.
+     */
+    stream leaf() noexcept (false);
```
ErlendHaa, line 76: We only need the eof state, not the entire handle. Would add `bool peof()` instead. The problem with a new stream of the underlying tell, is that if someone starts messing with it, they break the original stream. At the very least what you return here should be const.

python/dlisio/load.py
```diff
@@ -88,88 +88,209 @@ def load(path, error_handler = None):
...
+        In the process updates parsers knowledge about data eof and next logical
+        file offset.
+        """
+        ltell = self.stream.ltell
+        try:
+            self.stream.get(bytearray(1), ltell, 1)
+        except Exception as e:
+            self.error_handler.log(
+                core.error_severity.critical,
+                "dlis::load: Testing data availability",
```
ErlendHaa, line 254: This error message makes zero sense to me. Testing data availbillity? Error on opening new logical file? This function not open a new logical file?

python/dlisio/load.py
```diff
@@ -86,83 +88,245 @@ def load(path, error_handler = None):
...
+        """
+        sulsize = 80
+        sulbytes = bytearray(sulsize)
+        read = self.stream.get(
+            sulbytes, self.stream.ltell, sulsize)
+
+        if read != sulsize:
+            self.error_handler.log(
+                core.error_severity.major,
+                "dlis::load: Retrieving SUL",
```
ErlendHaa, line 324: We are reading the Sul right? Why not keep it simple?

python/dlisio/load.py
```diff
@@ -86,83 +88,245 @@ def load(path, error_handler = None):
...
+        Reports specification violations if detected.
+        """
+        expected = not self.sul and (len(self.lfs) == 0)
+        try:
+            core.findsul(self.stream, self.error_handler, expected)
+
+            if not expected:
+                self.error_handler.log(
+                    core.error_severity.minor,
+                    "dlis::load: Retrieving SUL",
```
ErlendHaa, line 289: Looking for sul ?

python/dlisio/load.py
```diff
@@ -86,83 +88,245 @@ def load(path, error_handler = None):
...
+                    "New SUL will be read",
+                    "")
+
+            return True
+        except RuntimeError as e:
+            if expected:
+                self.error_handler.log(
+                    core.error_severity.minor,
+                    "dlis::load: Retrieving SUL",
+                    "Exactly one SUL is expected at the start of the file: "
```
ErlendHaa, line 303: .. but found 0

lib/src/dlis/io.cpp
```diff
@@ -93,6 +101,48 @@ long long findsul( dlisio::stream& file ) noexcept (false) {
...
+        if (!expected) {
+            throw e;
+        }
+        // default maxread value comes from real files with trash before SUL
+        constexpr auto maxread = 1700;
+        file.seek(lfrom);
+        offset = findsul(file, maxread);
+
+        assert(offset > lfrom);
+        const auto debug = "SUL found at {} (dec), but expected at {}";
```
ErlendHaa, line 168: tell, ptell?

python/dlisio/load.py
achaikou: Decided on end_of_data

python/dlisio/load.py
```diff
@@ -88,88 +88,209 @@ def load(path, error_handler = None):
...
+        self.error_handler = error_handler
+        self.is_tif = is_tif
+        self.path = path
+
+        self.lfs = []
+        self.sul = None
+        self.stream = None
+
+        self.open_next_at = 0
+        self.data_eof = False
```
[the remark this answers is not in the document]
achaikou: Renamed to data_end, but I am against splitting this, because it means too many different things
- physical file EOF
- file truncated
- file truncated and padded which is discovered by findoffsets
- file not truncated but padded which is discovered by lfp

And all of them mean that we no longer have valid data to process.

python/dlisio/load.py
achaikou: 1. I feel like adding inline comments here spoil the flow too much, so I moved the explanation to logical_eof docs?
2. rp66 is called explicitly, stream_open name left as is because nothing better comes to mind :D 
3. eof stuff discussed

python/dlisio/load.py
```diff
@@ -86,83 +88,243 @@ def load(path, error_handler = None):
...
+        ltell = self.stream.ltell
+        try:
+            self.stream.get(bytearray(1), ltell, 1)
+        except Exception as e:
+            self.error_handler.log(
+                core.error_severity.critical,
+                "dlis::load: Testing logical eof",
+                e,
+                "",
+                "Load is suspended",
```
ErlendHaa, line 255: Indexing is suspended, not load

PR #334, opened 2021-03-03
LIS - Add tests for different file layouts w.r.t. tape marks and padding

ErlendHaa:


python/data/lis/layouts/README.rst
```diff
@@ -0,0 +1,48 @@
...
+            |-> File Header
+            |-> File Trailer
+            |-> File Header
+            |-> File Trailer
+
+That is, a single REEL, containing a single TAPE, containing 2 Logical Files.
+Each containing only a FHLR and a FTLR. All trailer Logical Records are
+present.
+
+Note that the actual Logical Records contain dummy data and can't be parsed by
```
achaikou, line 22: To discuss: apparently this might not be a good idea. It was difficult for me by looking at the files to figure out which record is which. Surprisingly real files give more tells about what's going on because their Headers/Trailers actually contain strings and are of more-or-less expected length. At least they were easier to investigate than custom ones.
So maybe we need to make this files more human-friendly.

achaikou (commented):
Testing TIF might be even more evil than implementing floats :smile:

python/tests/lis/test_physical_layout.py
```diff
@@ -0,0 +1,81 @@
...
+    assert_load_correctly(fpath)
+
+def test_broken_01():
+    fpath = 'data/lis/layouts/broken_01.lis'
+    assert_load_correctly(fpath)
+
+def test_broken_02():
+    fpath = 'data/lis/layouts/broken_02.lis'
+    assert_load_correctly(fpath)
+
```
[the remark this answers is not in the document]
ErlendHaa: 1. 2 TM(1) means EOF, hence we stop indexing. 1 TM means we continue, and hit the padding
2. No
3. In LIS the only defined padding is directly after a PR. In TIF I was pretty sure that double EOF TM meant EOF, hence nothing should follow
4. No, most common I would think is TM(1) after each LF. But note that if both reel and tape is present, these are considered _one_ LF, hence TM(1) only after the last record of the two

python/dlisio/lis/load.py
```diff
@@ -76,6 +76,11 @@ def read_as_tapemark(f):
         index = f.index_records()
         truncated = f.istruncated()
 
+        # If not a single record could be indexed, close the file and stop here
+        if index.size() == 0 and truncated:
```
achaikou, line 80: Why it has to be truncated then?

python/dlisio/lis/load.py
```diff
@@ -89,7 +94,7 @@ def read_as_tapemark(f):
         # All delimiter records are indexed separately by index_records. If
         # applicable the raw record is extracted and correctly stored. In any
         # case the filehandle is closed before we continue indexing the file.
-        first = index.explicits()[0]
+        first = index.explicits()[0] if len(index.explicits()) else None
         if is_delimiter(first):
```
achaikou, line 98: So now it can create a file with empty index because is_delimiter(None) will return false?

python/data/lis/layouts/README.rst
```diff
@@ -0,0 +1,56 @@
...
+any parsing routine. 
+
+================== ===========================================================
+Filename           Desciption
```
achaikou, line 26: Will update the files as per agreement as a follow-up of this PR :+1:

PR #335, opened 2021-03-03
Lis types

achaikou:


lib/include/dlisio/lis/types.h
```diff
@@ -43,16 +43,36 @@ const char* lis_i32(const char*, std::int32_t*);
...
- * middle */
+/* (reprc 70) 32-bit fixed point - 2's complement with binary point in the
+ * middle
+ *
+ * Note: some numbers can't be represented correctly by IEEE 754 floats.
+ * Problem happens for values with high precision which do not fit into floats.
+ * Using double type instead would have solved the issue.
+ * This type, however, is not expected to be used frequently as it's too
+ * different from the standard ones. Hence precision loss is considered
+ * acceptable.
```
ErlendHaa, line 74: On all these notes, maybe add some specifics on the precision loss. E.g. `Precision loss after 9th decimal` or something to that effect. And `values greater than float max, is capped at float max 3.4028235 × 10^38`

lib/test/lis/types.cpp
```diff
@@ -130,49 +131,312 @@ TEST_CASE("16-bit floating point", "[type]") {
...
+    SECTION(" maximum LIS-stored positive number (data loss)") {
+        /* E = 2^15 - 1, M = 2^15 - 1 */
+        const float expected =
+            (1 - std::pow(2.0f, -15)) * std::pow(2.0f, std::pow(2.0f, 15) - 1);
+        const bytes<4> input = { 0x7F, 0xFF, 0x7F, 0xFF };
+
+        float v;
+        lis_f32low( input, &v );
+        CHECK( v == expected );
+        // Failing: too big number, can't be represented by float
```
ErlendHaa, line 304: Why are these here?

lib/test/lis/types.cpp
```diff
@@ -130,49 +131,312 @@ TEST_CASE("16-bit floating point", "[type]") {
...
+
+            float v1;
+            lis_f32fix( input1, &v1 );
+            CHECK( v1 == expected1 );
+
+            float v2;
+            lis_f32fix( input2, &v2 );
+            CHECK( v2 == expected2 );
+
+            // value can't be represented by float as a different number
```
ErlendHaa, line 416: What do you mean? should this stay?

lib/test/lis/types.cpp
```diff
@@ -130,49 +131,312 @@ TEST_CASE("16-bit floating point", "[type]") {
...
+        float v;
+        lis_f32low( input, &v );
+        CHECK( v == expected );
+        // Failing: too big (modulo) number, can't be represented by float
+        // CHECK (v >= std::numeric_limits<float>::lowest());
+    }
+
+    SECTION(" near-smallest architecture-represented positive number ") {
+        /* E = -2^7, M = 1 */
+        // Note: even smaller number seems to be with E = -2^7 - 6
```
ErlendHaa, line 322: How is this helpful?

lib/test/lis/types.cpp
achaikou: Boundaries definition.
It's important to know the range of the data we do and do not support to have a clear picture.

I would have uncommented and marked the section as xfail if catch2 supported this option of marking sections as failed. Opted out for commented checks instead.

lib/test/lis/types.cpp
achaikou: Title and notes before all of that state that it's difficult to define the true and correct boundaries.
By my logic `E = -2^7, M = 1` should have been the smallest number.

It however appeared not to be, which shows that boundaries assumptions are not precise.

lib/test/lis/types.cpp
achaikou: That's the other wording for precision loss.
input1 and input2 are different numbers in the original type.
but after conversion they are different due to precision loss due to float limits.

lib/include/dlisio/lis/types.h
achaikou: Precision loss won't be very specific, it would be approximate.
But will try.

PR #338, opened 2021-03-08
Drop python 3.5 support and add python 3.9 to CI/CD

ErlendHaa:


.circleci/config.yml
```diff
@@ -202,6 +202,12 @@ jobs:
       - python_build:
           docs: "-DBUILD_DOC=ON -DSPHINX_ARGS=-WT"
 
+  python-39:
+    docker:
+      - image: circleci/python:3.5
```
achaikou, line 201: Fix that instead and add testing?

python/tests/dlis/test_curves_reprcodes.py
```diff
@@ -389,9 +389,9 @@ def test_ascii_broken_utf8():
 
 def test_datetime_invalid():
     fpath = 'data/chap4-7/iflr/invalid-dtime.dlis'
-    # run this test from python 3.6 only as code doesn't fail on python 3.5
-    # check might be removed once we remove support for python 3.5
-    if sys.version_info.major >= 3 and sys.version_info.minor > 5:
+    # run this test from python 3.6 through 3.8 only as code doesn't fail
```
achaikou, line 392: Try reverting it all back :D

PR #344, opened 2021-03-19
Identify multidimensional curves in DFRS and raise NotImplementedError

ErlendHaa:


lib/include/dlisio/lis/protocol.hpp
```diff
@@ -11,6 +11,16 @@
 
 namespace dlisio { namespace lis79 {
 
+/** Size of LIS type
+ *
+ * Return the size of bytes needed to store the LIS type. E.g. lis::f32 is a 4
+ * byte type. Variable length types such as lis::string and lis::mask return 0.
```
achaikou, line 17: I would prefer exception or -1?
0 seems like a valid value. Data not weights anything? Well, ok?
-1 gives clear indication something is wrong.

lib/include/dlisio/lis/protocol.hpp
```diff
@@ -11,6 +11,16 @@
 
 namespace dlisio { namespace lis79 {
 
+/** Size of LIS type
```
achaikou, line 14: 1. Why this function is in protocol.hpp, not in any typey-stuff?
2. Maybe more explicit clarification in name or in docs is needed to note that it's src (specification) type, not dst (your computer) type. One can "store LIS type" on both, so it's too vague.

lib/src/lis/protocol.cpp
```diff
@@ -35,6 +35,28 @@ constexpr const int lis::tape_trailer::size;
 constexpr const int lis::reel_header::size;
 constexpr const int lis::reel_trailer::size;
 
+int sizeof_type(int reprc) noexcept (false) {
+    using repcode = lis::representation_code;
+    const auto code = static_cast< repcode >(reprc);
+
+    switch( code ) {
```
achaikou, line 42: I am not fond of repeat of the part of the logic in dfs_fmtstr, but using this method there might reduce readability, so I leave the decision to you.

PR #348, opened 2021-03-22
Propagate indexing errors to python - and the enduser

ErlendHaa:


python/dlisio/ext/lis.cpp
```diff
@@ -365,9 +365,10 @@ void init_lis_extension(py::module_ &m) {
     ;
 
     py::class_< lis::record_index >( m, "lis_record_index" )
-        .def( "explicits", &lis::record_index::explicits )
-        .def( "implicits", &lis::record_index::implicits )
-        .def( "size",      &lis::record_index::size )
+        .def( "explicits",    &lis::record_index::explicits )
+        .def( "implicits",    &lis::record_index::implicits )
+        .def( "size",         &lis::record_index::size )
+        .def( "isincomplete", &lis::record_index::is_incomplete)
```
achaikou, line 370: why isincomplete and not is_incomplete?

lib/include/dlisio/lis/io.hpp
```diff
@@ -100,11 +101,16 @@ struct record_index {
      */
     bool is_incomplete() const noexcept (true);
 
+    /* Stores the error msg from the (if applicable) failed indexing process
```
achaikou, line 104: It doesn't store the message, it returns it.

Plus: Propegate -> Prop(a)gate in commit message

python/tests/lis/test_physical_layout.py
```diff
@@ -211,7 +211,7 @@ def test_truncated(filename, assert_error):
     fpath = 'data/lis/layouts/' + filename
     lf = Expected(records=1, ttlr=None, rtlr=None)
     assert_load_correctly(fpath, [lf])
-    assert_error("File likely truncated")
+    assert_error("Indexing failed")
 
 # TODO: all wrong file tests have same error message, may be precised
```
achaikou, line 216: Thanks :D This may be removed now :D

achaikou (commented):
Fix what you want to fix and merge

python/dlisio/ext/lis.cpp
ErlendHaa: Convension. In c++ it's common to use `is_`. e.g. [1], while in python it's common to to drop the underscore. e.g. [2]
[1] https://en.cppreference.com/w/cpp/types/is_same
[2] https://docs.python.org/3/library/functions.html#isinstance

PR #350, opened 2021-03-24
Add LIS curves placeholder tests

achaikou:
Initial black-box curves tests for currently prioritized features.
Behavior is not defined yet, files may contain mistakes.


python/tests/lis/test_curves.py
```diff
@@ -42,3 +44,544 @@ def test_read_curves():
...
+# DEPTH:
+# 1. placeholders. currently expected behavior may be changed on implementation
+# 2. more tests needed if depth can be negative
+# 3. might need test mode = 0, direction flag inconsistent with actual direction
+# if we decide to implement any checks
+# 4. Issues with inconsistent spacing and units: if depth units (14) != frame
+# spacing units (9), we might have problems with defining inconsistency
+# 5. Some information can be check or deduced if certain entries are missing.
+# Suggestion is to fail on all wrong or missing important information cases
+# until we get the real file with these situations.
```
ErlendHaa, line 282: Totally agree! Let's not implement _any_ value-deduction before we actually see a need for it. Great that you have comments everywhere that this would be a possibility 👍

ErlendHaa (approved):

PR #351, opened 2021-03-25
Support lis::string as channel type, multi-entry channels, suppressed channels and skipping fast channels

ErlendHaa:


lib/src/lis/protocol.cpp
```diff
@@ -487,6 +487,21 @@ std::string dfs_fmtstr( const dfsr& dfs ) noexcept (false) {
             throw std::runtime_error(fmt::format(msg, size, code, mnem));
         }
 
+        /* Suppressed output
+         *
+         * If the reserved_size in the Spec Block is negative, the bytes is
```
achaikou, line 492: the bytes (are)

python/dlisio/ext/lis.cpp
```diff
@@ -126,6 +126,9 @@ noexcept (false) {
             return;
         }
 
+        else if (*f == LIS_FMT_SUPPRESS) {
+            ptr++;
```
achaikou, line 130: Right now you have many "S" values as per 1 suppressed  channel, which is a bit unexpected.
Consider if you don't want them to behave as strings, so instead of `SSSSSSSS` have `S__8` to skip 8 bytes

lib/src/lis/protocol.cpp
```diff
@@ -487,6 +487,21 @@ std::string dfs_fmtstr( const dfsr& dfs ) noexcept (false) {
...
+         * reserved in the file, but the output is suppressed. I.e. should be
+         * ignored.
+         *
+         * This is communicated in the format-string with the fmt-code
+         * LIS_FMT_SUPPRESS.
+         *
+         */
+        if ( size < 0 ) {
```
achaikou, line 500: Add a note somewhere what takes precedence: suppressed channel or samples.

Right now if samples=11, reprcode_size = 4 and suppressed = -2, samples/repcode size will complain. That is as good behavior as any, but explicit note that we are aware of what's going on would be good.

python/dlisio/lis/curves.py
```diff
@@ -20,7 +20,7 @@
 }
 
 
-def curves(f, dfsr, strict=True):
+def curves(f, dfsr, strict=True, skip_fast=False):
```
achaikou, line 23: If feature is planned to be permanent, add some test case based on the most simple file?

lib/src/lis/protocol.cpp
```diff
@@ -479,15 +479,25 @@ std::string dfs_fmtstr( const dfsr& dfs ) noexcept (false) {
...
-        std::int16_t entries = size / s;
-        fmt.append( entries, f );
+
+        if ( f == LIS_FMT_STRING ) {
+            /* A lis::string does not encode its own length, hence we have to
+             * embed the length into the format-string. E.g. the format-string
+             * of a 256 character lis::string is "a256".
+             */
+            std::int16_t entries = std::abs( size );
+            fmt += (f + std::to_string( entries ));
```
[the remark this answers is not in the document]
ErlendHaa: I don't quite get what you mean about "prevent changing the fmt string".

I quite like the unbound length solution we have now, even tho it's unlikely to ever be a problem.

python/dlisio/lis/curves.py
```diff
@@ -124,9 +124,35 @@ def curves(f, dfsr, strict=True):
         alloc,
     )
 
+def spec_dtype(spec):
```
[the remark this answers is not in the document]
ErlendHaa: 1. Yeah, only thing not explicitly checked in fmtstr is samples, which we don't support atm. As the logic is likely to change a bit when we do, I'm not gonna bother

python/dlisio/ext/lis.cpp
```diff
@@ -110,14 +110,44 @@ void read_frame( const std::string& fmt,
...
-    assert_overflow(ptr, end, src_skip);
-    lis_packf(fmt.c_str(), ptr, dst);
-    dst += dst_skip;
-    ptr += src_skip;
+            auto str = py::str(ptr, len);
+            swap_pointer(str);
+            ptr += len;
+        }
+
+        else {
```
achaikou, line 145: might be on one line here and in other places

python/dlisio/lis/curves.py
```diff
@@ -44,6 +44,13 @@ def curves(f, dfsr, strict=True):
         this restriction and dlisio will append numerical values (i.e. 0, 1, 2
         ..) to the labels used for column-names in the returned array.
 
+    skip_fast : boolean, optional
+        By default (skip_fast=False) curves() will raise if the dfsr contain
```
achaikou, line 48: dfsr contain(s) 
curves can be called with only one dfsr

python/dlisio/lis/curves.py
```diff
@@ -138,6 +139,13 @@ def spec_dtype(spec):
             spec, spec.reserved_size, spec.samples, sample_size
         ))
 
+    # As strings does not have encoded length, the length is implicitly given
```
achaikou, line 148: strings (do)

PR #355, opened 2021-04-16
Introduce ErrorHandler to load

achaikou:
Note that this commit changes current default LIS load behavior.
Before load of broken files almost always returned correctly. All the errors were put in the log.
Now the same files will fail unless custom errorhandler is injected into the load function.

ErrorHandler is not passed down to the logical file yet, but there are plans to make it available at least for the curves.

Need to handle some warning messages might come in the future as well.


python/dlisio/lis/load.py
```diff
@@ -25,68 +24,117 @@ def load(path):
...
+                e,
+                "",
+                "Indexing stopped",
+                "Physical tell: {} (dec)".format(self.offset))
+            return False
+
+    def index_file(self):
+        """ Index current file.
+
+        Returns True if indexing was successful. Otherwise returns False or
```
ErlendHaa, line 125: Not super fan of this return true false thing. Feel free to ignore, but I propose:
```
while not indexer.isdone():
     try:
         indexer.index_file(): # Open and index
     except:
         indexer.close()
         raise

File: python/dlisio/lis/load.py [hunk 2]
```diff
@@ -96,52 +144,85 @@ def read_as_tapemark(f):
...
-                tape = HeaderTrailer(record)
-
-            elif first.type == core.lis_rectype.tape_trailer:
-                # The tape is already assigned to the LF's at this point, so
-                # just assign the trailer to that instance before initiating a
-                # new instance for the next tape.
-                tape.rawtrailer = record
-                tape = HeaderTrailer()
-
+                if index.size() != 1:
```
ErlendHaa, line 147: Style-comment that you can ignore. Maybe this becomes a bit more readable if we do:
```
try:
    if index.size() != 1:
        self.error_hander.log( ... )
 finally:
     self.file.close()
     
 try:
    record = self.file.read_record(first)
 except Exception as e:
     self.error_hander.log( ... )
 finally:
     self.file.close()

File: python/dlisio/lis/load.py [hunk 3]
```diff
@@ -25,68 +24,117 @@ def load(path):
...
-
-        # If not a single record could be indexed, close the file and stop here
-        if index.size() == 0 and index.isincomplete():
-            f.close()
-            break
+            self.error_handler.log(
+                core.error_severity.critical,
+                "dlisio.lis.load: file {}".format(self.path),
+                e,
+                "",
```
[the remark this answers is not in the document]
ErlendHaa: Ehm ne?
```
def log(self, severity='', context='', problem='', spec='', action='', debug=''):
     pass
```
Does not break the current interface. You also get the opertinity to make new caller code much cleaner, e.g:
```
issue = {
    'severity' : core.error_severity.critical,
    'context'  : 'dlisio.lis.load' ,
    'problem'  : exception_mgs,
    'action'   : 'stopped indexing "{}" at tell {}'.format(path, offset)
}
error_handler.log(**issue)
```

python/dlisio/lis/load.py
```diff
@@ -101,6 +98,7 @@ class FileIndexer:
     def __init__(self, path, error_handler):
         self.error_handler = error_handler
         self.path = path
+        self.complete = False
```
ErlendHaa, line 101: The attribute `self.file` on line 108 is no longer used, except within `index_logical_file`. There is no need to cache it on the class

ErlendHaa (commented):
LGTM 👍

python/dlisio/lis/load.py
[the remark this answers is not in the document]
achaikou: Ah, you meant to change everything...
Can't say that I like repeating all the argument names every time, but as it's cleaner in your opinion - changed.

python/dlisio/lis/load.py
```diff
@@ -42,6 +42,7 @@ def load(path, error_handler = None):
         try:
             indexer.index_logical_file()
         except:
+            indexer.close()
```
ErlendHaa, line 87: Why is this a separate commit?

python/dlisio/common/errorhandler.py
```diff
@@ -163,7 +163,7 @@ def __init__(self, info     = Actions.LOG_DEBUG,
         self.major    = major
         self.critical = critical
 
-    def log(self, severity, context, problem, spec, action, debug):
+    def log(self, severity, context='', problem='', spec='', action='', debug=''):
```
ErlendHaa, line 166: Commit msg: Errorhandler.load?

python/dlisio/lis/load.py
achaikou: It was supposed to be a separate commit because the first commit fails because I am not closing enough stuff (and I am no longer sure what, probably all). Then in second commit I refactor that thing. So in third commit I just wanted to explicitly fix the thing from first (probably) by closing everything. Might not be that.

But will merge commits 2 and 3.

PR #359, opened 2021-04-26
Pin fmtlib version in CI pipelines

ErlendHaa:
We keep running into issues when building fmtlib from source. To avoid
the hassle we switch to a stable version.


achaikou (approved):
Seems fine, but would be good to check we are not affected by cache on Windows CircleCI though, and GitHub says only like 3 checks were performed...

PR #362, opened 2021-05-04
Entries

achaikou:


lib/src/lis/protocol.cpp
```diff
@@ -294,6 +294,8 @@ const char* element( const char* xs,
                      lis::byte reprc,
                      lis::value_type& val )
 noexcept (false) {
+    if (lis::decay(size) == 0)
```
ErlendHaa, line 297: Do you mean for this to apply to component blocks too? In that case maybe update the commit msg. In any case, I think an early return in `read_entry_block` is "easier to follow", codewise

PR #363, opened 2021-05-06
Fix problems with buffer allocation in curves

achaikou:


python/dlisio/ext/dlis.cpp
```diff
@@ -650,14 +649,15 @@ noexcept (false) {
...
                          "Record is skipped", debug);
-        // we update the buffer as we go. Hence if error happened we need to
-        // go back and start rewriting updated data
-        dst = record_dst;
+        /* We update the buffer as we go. Hence if error happened we need to
+         * go back and start rewriting updated data.
+         * All the processed frames, regardless of the record they belong to,
+         * are preserved.
+         */
+        dst = static_cast< unsigned char* >(info.ptr) + (frames * itemsize);
```
ErlendHaa, line 657: 1) Would prefer to keep line 641, but rename to `dst_beg`, and here do `dst = dst_beg + (frames * itemsize).
2) Doing this in `handle` is conceptually weird, so we should move this line to after handle() in the try catch. Note that it's only when `read_fdata_record` fails we need to rewind the dst pointer. The other to cases/try's, doesn't touch it. I.e. move this line to after line 687
3) The comment needs a bit of love, or removal. It tells me a lot of semi-irrelevant stuff, but not what's actually going on.

> When failing to write a frame (row) to dst, the position of the dst pointer is left in an undefined state. I.e. it may be left anywhere within the partially written row. We discard the row completely by rewinding the dst pointer back to the start of the row.

4) likewise, I don't understand much of the commit msg (in either of the commits)

PR #366, opened 2021-05-10
Expand Process Indicators mask

ErlendHaa:
The property indicators mask in LIS Spec Block subtype 1 contains some
useful information. However it is rather cumbersome to work with as a
raw bit mask, particularly in Python, where masks are represented as
'bytes'.

The mask is expanded to a struct in C++, mostly because the mask
contains both int and boolean values. The struct is converted to a
python dict, to get a more ergonomic python interface.


achaikou (approved):
Look good!

PR #368, opened 2021-05-10
Introduce lis::io_device::read_records

achaikou:


ErlendHaa (approved):
Other than some newly introduced line break styles this looks good to me!

PR #376, opened 2021-05-20
Document backdoor access to units

achaikou:
While we struggle with creating good interface for units in current
design, units are actually already accessible to the users, though few
may know about it.

As this feature appears to be quite important, it makes sense to
properly document at least the "back door" access to it.

### DIFF
--- a/python/docs/dlis/metadata.rst
+++ b/python/docs/dlis/metadata.rst
@@ -92,6 +92,37 @@ to make it easy to work with programmatically:
    >>> channel.long_name
    'Bond Index'
 
+Units
+-----
+
+Alongside attribute values, a DLIS-file also contains a corresponding **units**
+field. This is true for all object attributes, even when units make little
+sense, such as for :attr:`Channel.long_name`.
+
+Currently there is no *nice* interface for accessing units with dlisio.
+They are, however, reachable through some of the more basic interfaces.
+Sensible units are also printed in :func:`BasicObject.describe`.
+
+Each metadata object has in itself an attribute attic,
+:attr:`BasicObject.attic`.
+The attic is a complete dict-like representation of the object, but without the
+syntactic sugar provided by the specialized class of the object.
+
+At the moment attribute units are only accessible through the attic:
+:code:`obj.attic['ATTRIBUTE'].units`.
+
+Every attribute has a *RP66V1 name* section in its docs that corresponds to
+the :code:`ATTRIBUTE` value, which must be used to retrieve the raw attribute.
+
+For example, given the object :code:`frame` and docs for :attr:`Frame.spacing`:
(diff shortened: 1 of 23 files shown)


python/docs/dlis/metadata.rst
```diff
@@ -92,6 +92,39 @@ to make it easy to work with programmatically:
...
+Units
+-----
+
+Many dlis attributes have units stored alongside the value, but, unfortunately,
+dlisio lacks official support for them.
+But while they are not accessible through any public interface, they still are
+reachable.
+
+.. warning::
+   Please note that further information discloses dlisio internals, thus is
```
ErlendHaa, line 104: Don't mention dlisio internals. The attic is public.
Would probably drop the entire warning tbh, we are still in pre v1.0.0 and everything can in theory be broken between versions (even `dlisio.load` :'D )

python/docs/dlis/metadata.rst
```diff
@@ -92,6 +92,39 @@ to make it easy to work with programmatically:
...
+   subject for a change and must be used with caution.
+
+.. note::
+   Please note that expected units are also printed in
```
ErlendHaa, line 108: Are we sure that `describe` is sufficiently covered to warrant this comment? E.g. `Measurement` ?

python/docs/dlis/metadata.rst
```diff
@@ -92,6 +92,39 @@ to make it easy to work with programmatically:
...
+   :func:`BasicObject.describe`.
+
+:attr:`BasicObject.attic` gives user access to the raw object as it is stored in
```
ErlendHaa, line 111: I'm unsure how familiar ppl are with BasicObject. maybe:

> On each object class (e.g. Frame, Tool, Parameter) there is an attribute `attic`, :attr:`dlisio.dlis.BasicObject.attic`, which is a more primitive dict-representation of the object. The attic is a complete dict-like representation of the object, but without the syntactic sugar provided by the specialised class it is a part of. At the moment attribute units are only accessible through this attic.

python/docs/dlis/metadata.rst
```diff
@@ -92,6 +92,37 @@ to make it easy to work with programmatically:
...
+the :code:`ATTRIBUTE` value, which must be used to retrieve the raw attribute.
+
+For example, given the object :code:`frame` and docs for :attr:`Frame.spacing`:
+
+.. code-block:: python
+
+   >>> frame.spacing
+   ... 800
+   >>> frame.attic['SPACING'].units
+   ... 0.5 ms
```
ErlendHaa, line 124: would this string be outputed as : `... '0.5 ms'` ?

ErlendHaa (approved):

PR #379, opened 2021-05-27
Support non-ASCII characters in filepath

achaikou:
On all modern platforms, but Windows prior to recent years, file-name
string passed to fopen is interpreted as UTF-8 [1].
Known and common workaround that suits to all Windows versions is to use
MultiByteToWideChar function.

[1] https://en.wikipedia.org/wiki/Unicode_in_Microsoft_Windows#UTF-8

### DIFF
--- a/lib/include/dlisio/file.hpp
+++ b/lib/include/dlisio/file.hpp
@@ -2,6 +2,7 @@
 #define DLISIO_FILE_HPP
 
 #include <cstdint>
+#include <cstdio>
 
 #include <lfp/lfp.h>
 
@@ -70,6 +71,9 @@ public:
 private:
     lfp_protocol* f;
 };
+
+/* Opens a file in 'rb' mode */
+std::FILE* fopen( const char* path ) noexcept (false);
 
 } // namespace dlisio
 
--- a/lib/include/dlisio/file.hpp
+++ b/lib/include/dlisio/file.hpp
@@ -1,5 +1,5 @@
-#ifndef DLISIO_STREAM_HPP
-#define DLISIO_STREAM_HPP
+#ifndef DLISIO_FILE_HPP
+#define DLISIO_FILE_HPP
 
 #include <cstdint>
 
@@ -73,4 +73,4 @@ private:
 
 } // namespace dlisio
 
-#endif // DLISIO_STREAM_HPP
+#endif // DLISIO_FILE_HPP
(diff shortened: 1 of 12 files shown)


lib/src/sysutils.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <dlisio/sysutils.hpp>
+
+#ifdef _WIN32
+    #include <windows.h>
+#endif
+
+namespace dlisio {
+
+std::FILE* sysopen(const std::string& path) noexcept (false) {
```
ErlendHaa, line 14: Would prefer just `open`, and same for the file names. Alternatively, put it in stream.hpp to avoid file clutter

lib/src/sysutils.cpp
```diff
@@ -0,0 +1,42 @@
...
+    if (wpathlen <= 0) {
+        const auto msg = "dlisio::sysopen: unable to count wpath length";
+        throw std::runtime_error(msg);
+    }
+
+    std::wstring wpath;
+    wpath.resize(wpathlen);
+    auto written =
+        MultiByteToWideChar(CP_UTF8, 0, path.c_str(), -1,
+                            const_cast<wchar_t*>(wpath.c_str()), wpathlen);
```
ErlendHaa, line 32: Drunken monkey? :D

ErlendHaa, line 32: Pretty sure you should use `wpath.data()` here [1] ? Or as Microsoft do: `path[0]` [2]

[1] https://en.cppreference.com/w/cpp/string/basic_string
[2] https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-multibytetowidechar#code-example

lib/include/dlisio/file.hpp
```diff
@@ -71,6 +73,9 @@ class stream {
     lfp_protocol* f;
 };
 
+/* Opens a file in 'rb' mode */
+std::FILE* fopen( const char* path ) noexcept (false);
```
ErlendHaa, line 77: Include mode to?

lib/include/dlisio/file.hpp
```diff
@@ -2,6 +2,8 @@
 #define DLISIO_FILE_HPP
 
 #include <cstdint>
+#include <cstdio>
+#include <string>
```
ErlendHaa, line 6: Do you need string?

PR #394, opened 2021-09-16
(Investigation) Adjust for scikit-build 0.12 fixes 

achaikou:
Scikit-build fixed some platform-related bug in version 0.12
https://scikit-build.readthedocs.io/en/latest/changes.html#scikit-build-0-12-0

> Architecture now always passed through -A to MSVC generators


achaikou:
Consider creating an issue for skbuild.

PR #411, opened 2022-03-09
Failure investigation hints to users and developers

achaikou:


python/docs/dlis/errors.rst
```diff
@@ -0,0 +1,179 @@
...
+.. note::  Please note that error messages might get updated while this sections
+           might not be, so treat them more as a hint than as a strict
+           guideline.
+
+.. contents:: :local:
+
+Errors on load
+--------------
+Load errors are in general non-recoverable and prevent user from further
+reading the file.
```
ErlendHaa, line 20: 1) Maybe start by mentioning that if load fails, it's most likely because the file is broken somehow. I don't think load will ever fail on a correctly written file? 

2) Mention errorhandler, as a potential remedy to the problem, and how it will read up-until the point of failure. Which makes it useful when the file is ill-formed at the end, but gives you no data when the file is ill-formed at the start 

3) A lot, if not all of the following errors are cause by the file being illformed in one of the following ways:
 - Trash at the start or end of the file
 -  Missing data in the middle
 - Truncated
 -  Zeroed out 

Which of the following errors that hits you, is dependent on exactly where in the file the above happens. I.e. the following errors are different manifestations of the same set of issues.

python/docs/dlis/errors.rst
```diff
@@ -0,0 +1,179 @@
...
+  field 'DEPTH.1.0' occurs more than once
+
+Happens when accessing curves. According to specification, every channel may
+appear in the frame once and only once.
+So this error may indicate:
+
+* Something is broken in frame-channel relationships and you should be cautious
+  with trusting the data
+* Channel is repeated for user convenience. It's likely the case if repeated
+  channel can be assumed to be index of the frame, like for example "DEPTH"
```
ErlendHaa, line 131: The latter point is a relatively common one, so maybe put that one first?

python/docs/dlis/errors.rst
```diff
@@ -0,0 +1,179 @@
...
+
+  frame.curves(strict=False)
+
+fmtstr would read past end
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. code-block:: text
+
+  corrupted record: fmtstr would read past end
+
+Error happens when reading curves. It indicates that more curves data is
```
ErlendHaa, line 143: > Error happens when reading curves.

If you got the error, you already know this. I would remove. Same for the error above

python/docs/dlis/errors.rst
```diff
@@ -0,0 +1,179 @@
...
+expected than there is actually present. dlisio doesn't know what precisely
+caused data to be incorrect, hence can't recover from it.
```
ErlendHaa, line 145: > It indicates that more curves data is expected than there is actually present. dlisio doesn't know what precisely caused data to be incorrect, hence can't recover from it.

Or: 

There is less data than what the metadata claims there to be. Either the metadata is incorrect or the file is missing data, or both. It's not obvious which one it is. The inconsistency between the metadata and the raw data makes it impossible for dlisio to correctly parse the data. Hence there is no recovery from this.

python/docs/dlis/investigation.rst
```diff
@@ -0,0 +1,171 @@
+Investigation hints
```
ErlendHaa, line 1: Manually debug broken files ?

python/docs/dlis/investigation.rst
```diff
@@ -0,0 +1,171 @@
...
+-------------
+
+You:
+
+ 1. Are in extreme need of reading as much data as possible from That One File
+    which dlisio fails to read, or need to investigate it
+ 2. Are very familiar with rp66 specification
+ 3. Have a good grasp on working with binary files
+ 4. Can build dlisio_ and lfp_ from source code and have enough knowledge to
+    debug it
```
ErlendHaa, line 26: Would condense this to something like:

> The nature of a DLIS-file means that even a few incorrectly written bytes makes the file unparsable. From our experience, many files can be fixed by manual inspection, and correction of just a handful of bytes. Or by removing garbage at the end of the file. For someone with experience with working with binary files and that is familiar with the RP66 specification, it might be feasible to fix broken files manually. This section tries to give some hints to how you would go about fixing broken files

python/docs/dlis/investigation.rst
```diff
@@ -0,0 +1,171 @@
...
+*probably* means that *data* itself is fine and is written as intended. Errors
+on accessing objects indicate that intention didn't match RP66 specification.
+
+
+Visual inspection
+-----------------
+
+Visually inspect the file to decide if failure cause is obvious.
+
+Open the file in binary editor, scroll through it.
```
ErlendHaa, line 46: .. in _a_ binary editor _and_ scroll ..

python/docs/dlis/investigation.rst
```diff
@@ -0,0 +1,171 @@
...
+
+* Make sure that it looks like every other dlis file that opens correctly.
+  Regardless of the extension, file contents might not be dlis.
```
ErlendHaa, line 49: Make sure it is a dlis file. Often the extensions lie. Look for the SUL or the first VR header

python/docs/dlis/investigation.rst
```diff
@@ -0,0 +1,171 @@
...
+
+Figure out failure reason
+-------------------------
+
+Results of your investigation might be something that makes sense - certain
+specification violation, missing bytes or some specific but consistent between
+several files error.
+
+Sometimes, however, it won't be possible to figure out what exactly went wrong.
+Then failure can't be classified and there is little that can be fixed.
```
ErlendHaa, line 100: 5 lines to tell you the obvious thing: "if you can't find the cause, you can't fix it"? Would remove all together

python/docs/dlis/investigation.rst
```diff
@@ -0,0 +1,171 @@
...
+Fixes
+-----
+In certain cases you might be able to fix the file.
+
+Remember that there is no guarantee that your fix will lead to the correct file:
+
+* fix itself might be wrong
+* there might already be too much data missing
+* fix might create screwed image of the file
+* there might be more errors further down the file
```
ErlendHaa, line 151: Would remove this too. It's all very obvious? I don't see how this adds any value

ErlendHaa (commented):
Nice work!

python/docs/dlis/investigation.rst
ErlendHaa: The condense it a bit? Something like this should capture it?

> We've seen files where there is a lot of incorrectly written bytes and no obvious pattern to it. In such cases, the effort of fixing them drastically increases

python/docs/dlis/errors.rst
ErlendHaa: 1. maybe something like:
> Allthough there are still some features of the DLIS spec that dlisio doesn't implement full support for, it should be able to load all dlis files that conform to the spec. From experience we know that there exists a lot of files out there that doesn't strictly adhere to the spec, and these might case load to fail. It's rather common that files are truncated, the file is zeroed out from some point or that there are some extra "trash" bytes somewhere in the file. All of which can through dlisio parsing rutines off. 
2: 👍 
3: 👍

ErlendHaa (approved):
lgtm

PR #427, opened 2023-10-02
upated links to energistics for rp66v1 and lis-79

geolr:
failed for me when searching, found at https://www.energistics.org/hosted-standards


sveinung-r:
I notice also that there are several other links that needs to be updated. Would you like to do this and fix the links you provided or should we close this PR and we will look into it when we have time?

geolr:
😱 sorry for the fail...
probably some more eyeballs needed. 
https://www.energistics.org/hosted-standards should be source for urls
Let me try again, then create a new PR

PR #428, opened 2023-10-03
fixed urls in readme

geolr:
now, no dead links should be left

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ behalf when such violations pose any ambiguity.
 ## Installation ##
 
 dlisio supplies pre-built python wheels for a variety of platforms and
-architectures. The wheels are hosted through [PyPi](https://pip.pypi.io) and
+architectures. The wheels are hosted through [Python Package Index PyPi](https://pip.pypa.io) and
 can be installed with:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@
 
 dlisio is an LGPL licensed library for reading well logs in Digital Log
 Interchange Standard (DLIS V1), also known as [RP66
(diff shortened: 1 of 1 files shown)


achaikou:
Hi, thank you for your discoveries and contribution, but I think we need to fix more before this change could be merged.

- As you might have noticed, there is no actual specification under the RP66 link. It is empty, thus meaningless. I think we have to wait until site owners actually put it there before doing anything about it
- Sveinung meant that both specification links are used across the project. We would anyway have to fix other ones, not just the ones in the Readme. It would be good to do it everywhere at the same time.
- As you see, something is broken in the CI, and without it nothing could be merged. We would need to allocate time to check what is going on.
- We do not have contributing guidelines written anywhere, but internally there are some rules about what we want the commit history to look like

Please also note that dlisio is not actively worked on for the moment and we are on limited capacity for efficient managing of the project. So if you'd like to fix those issues yourself it might take significant time from our side to respond.

geolr:
True, the RP66 V1 is not really there either...
Sent mail to energistics (you on CC).

Should I close this then, or do you want to keep the other links?

achaikou:
Wow, thank you very much for doing this! :heart_eyes: 

If you want to have it merged, I recommend you to do the following:

Have just one commit. Rename it to
`Fix broken links in Readme`
(we use [this](https://cbea.ms/git-commit/) guideline)

This message implies that only some Readme links are fixed and allows us in the future to fix remaining dead links across the repo without forcing you to spend time on it now.

I will try to find time next week to investigate tests failures so we can have it merged rather sooner than later, but no promises.
(Changes of course should have no impact on the code, but it is a bit weird to merge with failing CI...)

If you don't want to deal with it yourself, you can close the issue, and I will add you as a co-author whenever we get to fix it.

PR #432, opened 2023-11-14
Update broken links

achaikou:


README.md
```diff
@@ -39,8 +39,8 @@ behalf when such violations pose any ambiguity.
 ## Installation ##
 
 dlisio supplies pre-built python wheels for a variety of platforms and
-architectures. The wheels are hosted through [PyPi](https://pip.pypi.io) and
-can be installed with:
+architectures. The wheels are hosted through [Python Package Index
+PyPI](https://pypi.org/) and can be installed with:
```
ajaust, line 43: This more a style suggestion. Feel free to ignore the suggestion.
```suggestion
architectures. The wheels are hosted through the [Python Package Index
(PyPi)](https://pypi.org/) and can be installed with:
```

ajaust (approved):
Looks good to me overall. I only added one optional comment.

achaikou:
Thanks, @geolr !
RP66v1 specification is available now, so links are updated.

nicolascordoba1:
In the documentation there is a citation to Appendix G but I'm not seeing it on the right side menu

ajaust:
I didn't find any part of the documentation where "Appendix G" is mentioned. The only place seems to be a comment in the source code. The source code refers to the [Custer Tape Subset](https://archive.org/details/S10046/mode/2up) which has an Appendix G. If you think that there is something actually wrong with the documentation, please don't hesitate to open an issue.

PR #437, opened 2024-07-26
Numpy 2.0

achaikou:
Planning to test wheels before making a release, but think it should be ok.

Update:
1. Setuptools just removed test command https://github.com/pypa/setuptools/commit/4c0b9f3ee6ee47c597572655567f215c08c90137
Change is in 72.0, now 70.1 or something near it is picked up, so surprise might yet to come
2. pytest-runner is deprecated. Will have to deal with it somehow
https://pytest-runner.readthedocs.io/en/latest/index.html
3. Invoking with python setup.py is deprecated, which is a problem, need to figure out what to do now.
https://packaging.python.org/en/latest/discussions/setup-py-deprecated/
Last two points rather not in this PR.

Update 2:
- flow that is affected by setuptools test command is not really in CI (it would be discovered by bigendian job, if setuptools there were updated)
- yes, we fail with newest setuptools (though there is already a version where test command is restored). That is about `Unknown distribution option: 'tests_require'`, but I am not sure if it is our test_require problem or if it is skbuild problem or if it is still setuptools didn't revert something back. I think that even if I don't pin setuptools, nothing should fail here (bigendian anyway using old version). But the issue should be addressed later anyway.

Update 3:
Seems like it is our test-require that breaks setuptools. If whole "remove python setup.py test" thing is done, looks like new setuptools works fine.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -30,9 +30,9 @@ jobs:
             arch: aarch64
           - os: ubuntu-20.04
             arch: i686
-          - os: macos-11
+          - os: macos-13
             arch: x86_64
-          - os: macos-13-xlarge
+          - os: macos-14
             arch: arm64
 
     steps:
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -40,14 +40,14 @@ jobs:
         run: |
             git config --global core.autocrlf false
 
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       # Used to host cibuildwheel
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: '3.10'
 
-      - uses: docker/setup-qemu-action@v2
+      - uses: docker/setup-qemu-action@v3
         if: ${{ matrix.arch == 'aarch64' }}
         name: Set up QEMU
 
@@ -83,7 +83,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - uses: actions/download-artifact@v3
 
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -29,7 +29,7 @@ jobs:
           - os: ubuntu-20.04
             arch: aarch64
           - os: ubuntu-20.04
-            arch: i686
+            arch: i686 #builds from cibuildwheel, qemu not needed
           - os: macos-13
             arch: x86_64
           - os: macos-14
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -68,7 +68,8 @@ jobs:
           # musllinux arch skip: unlikely someone wants to run dlisio on alpine on non-standard architecture
           # musllinux cp38 skip: latest available numpy doesn't provide musslinux wheels, so it is unlikely useful
           # macosx arm 38 skip: cibuildwheel can't test it
-          CIBW_SKIP: pp* *-musllinux_i686 *-musllinux_aarch64 cp38*-musllinux* cp38*-macosx_*arm64 cp36-* cp37-*
+          # 36, 37, 313 skip: python versions not supported
+          CIBW_SKIP: pp* *-musllinux_i686 *-musllinux_aarch64 cp38*-musllinux* cp38*-macosx_*arm64 cp36-* cp37-* cp313-*
 
         run: |
             python -m cibuildwheel --output-dir wheelhouse python/
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -67,8 +67,8 @@ jobs:
           CIBW_ARCHS: ${{ matrix.arch }}
           # musllinux arch skip: unlikely someone wants to run dlisio on alpine on non-standard architecture
           # musllinux cp38 skip: latest available numpy doesn't provide musslinux wheels, so it is unlikely useful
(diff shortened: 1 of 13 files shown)


python/setup.py
```diff
@@ -2,7 +2,6 @@
 
 import os
 import skbuild
```
sveinung-r, line 4: About the commit message: I thought we concluded that skbuild only brought setuptoools.command.test into the symbol table by importing it and never overwrote it?

sveinung-r (changes requested):

python/requirements-dev.txt
```diff
@@ -1,6 +1,6 @@
 
 bandit
-numpy<2.0
```
ajaust, line 3: Would it make sense to pin the NumPy version to `>=2,<3`?

1. Pinning would make clear that future changes to `dlisio` should follow the new NumPy conventions. 
2. With the name of this PR I would expect that we want NumPy 2.0 now.

python/setup.py
```diff
@@ -36,7 +36,7 @@ def get_long_description():
     packages = ['dlisio', 'dlisio.dlis', 'dlisio.lis', 'dlisio.common', 'dlisio.dlis.utils'],
     license = 'LGPL-3.0',
     platforms = 'any',
-    install_requires = ['numpy < 2.0'],
+    install_requires = ['numpy'],
```
ajaust, line 38: If we want to pin Numpy to v2, this line should pin the version as well.

ajaust (commented):
Looks good to me, I just had a minor comment so I understand better how we should handle major version upgrade.

python/requirements-dev.txt
sveinung-r: We should not pin it to version 2 as long as it works with version 1, and we don't have any compelling reason to brak numpy 1 compatibility. This may cause unnecessary dependency conflicts for users that still have a reason to keep using dumpy 1.

When it comes to setting an upper bound for the version either way has its pros and cons. I prefer not to pin unless we discover that we need to.

python/setup.py
achaikou: We concluded that it now doesn't override setuptools, but they before supplied their own custom test command.
https://github.com/scikit-build/scikit-build/blob/ff9cde427e2ed8f6f683139a23286c063d510352/skbuild/command/test.py

sveinung-r: Yes, but they did not overwrite setuptoools.command.test in the symbol table, so you would still access setuptools test command when referring directly to setuptools.

achaikou: Yes, agree.
But which exact wording in the commit message you don't like then?

sveinung-r: "(It is though unclear if test command from setuptools was ever called by
any parts of our solution.)" and "However setuptools.command.test was
imported directly in the skbuild, where they exchanged 'test' with their
own command. So we happened to use it indirectly."

achaikou: 1. "(It is though unclear if test command from setuptools was ever called by
any parts of our solution.)" 

It was imported, but was it ever called? Did we just overrode this stuff because we didn't like it and was it ever part of our test execution?
Because I would assume that if us calling test either through skbuild or setuptools was somehow important, then something would have broke down once we removed the override line because some code that is the essence of the setuptools..test method would never be called.

2. I don't understand what is wrong
- setuptools.command test was imported directly in the skbuild by using from setuptools.command.test import test as _test
- it was done in place where they exchanged test with their own command - aka in the skbuild.command.test file
- we happened to use that import (of setuptools.command.test.test) indirectly because we never imported it, yet skbuild did.

sveinung-r: 1. Yes, I do think the setuptools test command was used when running our tests. I think skbuild test command was changed so that It behaves the same as the setuptools test command.

2. I don't think they exchange the setuptools test command. They just define their own. I may have misunderstood your wording here though. I read it as "we were using skbuild test command indirectly", but I think what you ment is "we use the import statement indirectly"?

achaikou: 1. Setuptools.test is deprecated, that's why I doubt it is used any longer. And I am not sure it was used because deprecation issue is from 2019.  Because of this uncertainty I added this comment as I feel it is important note.
2. Yes, by exchanged I didn't mean "go inside their code and substitute theirs with custom", but "we have extended their version and exchanged the original call with ours through `kw["cmdclass"]`, so that whenever something in setuptools calls test, our test will be called".
And yes, I mean that we use import statement indirectly.


I suggest you to propose your own wording here, because otherwise we will spend another 10 comments with me trying to figure out how to word that so it pleases you :smile:

sveinung-r: I can not see any reason why setuptools test command would not be called, so I think it was.

I don't think it is correct to say that anything in setuptools is replaced, rather that skbuild is used in place of setuptools (the setup() function from skbuild is called).

You can have a look at the wording in my PR to segyio (https://github.com/equinor/segyio/pull/576) and see if you agree with it.

.github/workflows/wheels.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
           # musllinux arch skip: unlikely someone wants to run dlisio on alpine on non-standard architecture
           # musllinux cp38 skip: latest available numpy doesn't provide musslinux wheels, so it is unlikely useful
           # macosx 38 skip: cibuildwheel can't test it
-          CIBW_SKIP: pp* *-musllinux_i686 *-musllinux_aarch64 cp38*-musllinux* cp38*-macosx_*:arm64 cp36-* cp37-*
+          CIBW_SKIP: pp* *-musllinux_i686 *-musllinux_aarch64 cp38*-musllinux* cp38*-macosx_*arm64 cp36-* cp37-*
```
sveinung-r: Are we not building and testing on MacOS arm64?

sveinung-r (approved):

.github/workflows/wheels.yaml
achaikou: Only for python 3.8: as commend above says, "cibuildwheel can't test it"

PR #443, opened 2025-05-23
Add reference to dliswriter in README

ajaust:
As it was discussed, this PR adds some references to dliswriter and their helper code to use dlisio in conjunction with dliswriter to create/write DLIS files.


achaikou (commented):
Rebase to pass the tests and can be merged :)

I still think that

> has very limited capability to write files

is a lie as we have limited capability to write data, but no capability to write files. Users are likely interested in the later.

ajaust:
I just rebased and pushed again. I also added a fixup to change the phrasing to "no capabilities to write files".

achaikou (approved):
Apply fixups and merge :)

PR #444, opened 2025-05-23
Update GitHub runners

ajaust:
GitHub's Ubuntu 20.04 runners have been deprecated since 2025-05-14 [1]. The runners are updated to use the `ubuntu-latest` image. The only exception is the `aarch64` build on Linux that uses `ubuntu-24.04-arm` which an `aarch64` runner which allows to build the wheels without emulation speeding up the build process.

[1]: https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -16,13 +16,13 @@ jobs:
     strategy:
       matrix:
         include:
-          - os: windows-2019
+          - os: windows-2022
             arch: AMD64
-            cmake_generator: "Visual Studio 16 2019"
+            cmake_generator: "Visual Studio 17 2022"
             cmake_generator_platform: "x64"
-          - os: windows-2019
+          - os: windows-2022
             arch: x86
-            cmake_generator: "Visual Studio 16 2019"
+            cmake_generator: "Visual Studio 17 2022"
             cmake_generator_platform: "Win32"
           - os: ubuntu-latest
             arch: x86_64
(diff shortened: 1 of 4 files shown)


.github/workflows/wheels.yaml
```diff
@@ -24,11 +24,11 @@ jobs:
             arch: x86
             cmake_generator: "Visual Studio 16 2019"
             cmake_generator_platform: "Win32"
-          - os: ubuntu-20.04
+          - os: ubuntu-latest
             arch: x86_64
-          - os: ubuntu-20.04
+          - os: ubuntu-24.04-arm
```
achaikou, line 29:
```
The only
exception is the `aarch64` build on Linux that uses `ubuntu-24.04-arm`
which an `aarch64` runner which allows to build the wheels without
emulation speeding up the build process
```

So either
1) add to commit message information that removing qemu will be done later, or
2) remove QEMU now, or
3) just simply switch to ubuntu-latest and lets deal with it all later.

Note that ubuntu-arm images are STILL in preview, so updating to it is questionable without additional reasons.

-----
_off-topic_
Apparently I did that for segyio just 4 months ago and I had **no** memory of it at all  :sweat_smile:

achaikou (commented):
Do windows ones too in order to not go back to this in a month?
`https://github.com/actions/runner-images/issues/12045`

As for macos, maybe lets not touch them for now.
Point was that one (macos14) is supposed to be ARM-runner and another (macos13) Intel.
So probably macos-14 we can switch to macos-latest in the future. Don't know what we would do with intel one, as in more recent releases they only are available as "large" runners.
We now build wheels on them, but I have no clue if
1) macos users still use Intel machines
2) cbuildwheel will be able to cross-compile for different macos architectures
3) we would be able to test that everything is working.

Looks like problem for later! :see_no_evil:

